PR #12, opened 2024-07-09
DCMAW-8264: Add logging framework and implement for the token lambda

sandy-may:
DCMAW-8264

### What changed
Adds a logger for use in lambdas. This uses AWS Powertools logger: https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/

This includes functionality for registering log messages and ensuring that the lambda can only have access to the registered log messages for that lambda using the type system. 

Currently there are limited properties in the log message - this can be expanded over time as and when the requirements surface.

This also adds a new ErrorOrValue type, superseeding the LogOrValue type. The rationale for this is that it feels cleaner to implement logs at the lambda level and not couple downstream services with the Logging framework.

For awareness - AWS have native ability now to write structured logs without the need for a library such as AWS Powertools. This could reduce the build size and need the dependency -> https://aws.amazon.com/about-aws/whats-new/2024/07/aws-lambda-search-filter-aggregate-function-logs/. I'd personally suggest merging as-is and I can work on analysing this today.

### Why did it change
To improve visibility of incoming requests that do not meet the required schema/spec and potential defects/server issues within the code.

## Checklists

- [x] No PII data logged. [See guidance here]
- [x] Documentation updated


src/functions/services/logging/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+There are two logs that do not follow this pattern - the STARTED and COMPLETED logs. These logs are useful to analyse whether a lambda has finished executing successfully. 
+
+Careful considering should be given if other logs that do not follow this pattern are added.
+
+## Regsistration of logs:
+
+Logs should be registered within the folder for the given lambda. See here for an example of how to create a log file following the required typing:
+
+[Registered common logs] (./commonRegisteredLogs.ts)
```
sandy-may, line 15: still trying to get this working

sandy-may: done!

jamesamooney (changes requested):
Appreciate the `errorOrValue` file isn't actually in this PR because we stole it for the previous PR but I wondered whether it's worth having a unit test for the function?

src/functions/services/logging/logger.test.ts
```diff
@@ -0,0 +1,125 @@
+import { LogAttributes } from "@aws-lambda-powertools/logger/lib/cjs/types/Log";
+import { ILoggerAdapter, Logger } from "./logger";
+import { LogMessage, RegisteredLogMessages } from "./types";
+import { Context } from "aws-lambda";
+
+describe("Logger", () => {
+  describe("Given there is a message to log", () => {
+    it("Writes a log with registered log data ", () => {
+      const loggingAdapter = new MockLoggingAdapter();
+      const mockLogger = new Logger<MockMessage>(
```
jamesamooney, line 10: Tiny tiny nut but why is this const `mockLogger`? Isn't this the real logger so we could call it logger? No value provided here so happy if you want to dismiss. If you want to change then there's other occurrences, I'm just going to comment once

src/functions/services/logging/logger.test.ts
```diff
@@ -0,0 +1,125 @@
+import { LogAttributes } from "@aws-lambda-powertools/logger/lib/cjs/types/Log";
+import { ILoggerAdapter, Logger } from "./logger";
+import { LogMessage, RegisteredLogMessages } from "./types";
+import { Context } from "aws-lambda";
+
+describe("Logger", () => {
```
jamesamooney, line 6: Haven't got to the logger file itself yet but was expecting to see a test to say if we try and log a message that isn't registered it throws an error. Do we want this functionality/test in place or a future enhancement, or have I misunderstood the value of registered log messages?

sandy-may: It shouldn't be possible given the type system. Give it a go!

src/functions/services/logging/logger.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.logger.info(
+      {
+        message,
+        messageCode: this.registeredLogs[message].messageCode,
+      },
+      data,
+    );
+  };
+
+  addContext = (context: Context): void => {
```
jamesamooney, line 32: Is logging `context` a nice to have or has `context` has ever/will ever provide us with useful information?

I'm not suggesting we remove it just keen to know more about it relevant to this project.

sandy-may: It adds this which I think is very useful:
image

src/functions/services/logging/logger.test.ts
sandy-may: I think this is a good shout. Updated

src/functions/services/logging/logger.test.ts
jamesamooney: Nice I like it. I still wonder whether there should be a unit test to show it won't work.. might be overkill but my thinking:
- our unit tests are very good documentation of what things should be doing
- Edge cases or unexpected inputs that TS isn't protecting against
- If for any reason we did create a runtime error (which I think is highly unlikely) the tests will catch this

Appreciate typescript will throw a type error when writing the test but I found at least one way to get around it:

 it('should throw an error when logging an unregistered message', () => {
        // @ts-expect-error: Testing a case that should not be allowed by TypeScript
        expect(() => logger.log('info', 'This should fail!!')).toThrow('Message type info is not registered');
        
Happy tho if you think this is overkill or if a logic change seems easier

sandy-may: Thanks @jamesamooney. Explicitly overriding typescript's inbuilt linting isn't a valid scenario for me. 

For quality here, we use a combination of types and unit tests. 

I'm happy to close this without changing.

jamesamooney (changes requested):
In our BE we have a file that documents what each message means, here is an example:

```
 {
    messageName: 'STARTED',
    message: 'lambda handler processing started',
    severity: Severity.Informational,
    userImpact: 'No impact',
    threshold: 'See Status thresholds in \'https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3351281665/DCMAW+-+Logging+Metrics+Alerts+-+Thresholds',
    logType: 'Status'
  },
```

Do we want similar in this repo? Appreciate the messages are quite simple right now but might be worth creating the file so it's not missed in future?

sandy-may:
> In our BE we have a file that documents what each message means, here is an example:
> 
> ```
>  {
>     messageName: 'STARTED',
>     message: 'lambda handler processing started',
>     severity: Severity.Informational,
>     userImpact: 'XXX',
>     threshold: 'XXX',
>     logType: 'Status'
>   },
> ```
> 
> Do we want similar in this repo? Appreciate the messages are quite simple right now but might be worth creating the file so it's not missed in future?

Hey @jamesamooney. good question. I think for now we should look at incremental changes. We already have a common registered log file so we have the functionality to expand this with relative ease.

There's no file bespoke to an individual lambda as I don't think we have a scenario (yet) that would need an individual registered log file. I've included in the README.md guidance for how to do this as and when we need it.

Happy to have a quick call about this if needed

jamesamooney (approved):

sandy-may:
> Appreciate the `errorOrValue` file isn't actually in this PR because we stole it for the previous PR but I wondered whether it's worth having a unit test for the function?

Yes, we should.

src/functions/types/tests/errorOrValue.test.ts
```diff
@@ -0,0 +1,18 @@
+import { errorResponse, successResponse } from "../errorOrValue";
+
+describe("Error or value", () => {
```
jamesamooney, line 3: Thanks!

jamesamooney (dismissed):

jamesamooney (approved):

PR #15, opened 2024-07-10
DCMAW-8264: WIP - Add validation to check request redirect_uri and registered redirect_uri

J-son1:
### What changed
- Add validation to check the stored/registered `redirect_uri` is valid when a request is made to the `/async/token` endpoint and the `/async/credential` endpoint.
- Update `validate()` method on `ClientCredentialsService` to validate the registered `redirect_uri` and update method name to reflect that the validation is being done on the `/async/token` endpoint request data
- Add method to the `ClientCredentialsService` to validate the supplied `redirect_uri` from the `/async/credential` request data

## Checklists

### DIFF
--- a/src/functions/asyncCredential/asyncCredentialHandler.test.ts
+++ b/src/functions/asyncCredential/asyncCredentialHandler.test.ts
@@ -812,16 +812,22 @@ class MockTokenSeviceValidSignature implements IVerifyTokenSignature {
 class MockFailingClientCredentialsServiceGetClientCredentialsById
   implements IClientCredentialsService
 {
+  validateTokenRequest(): ErrorOrSuccess<null> {
+    return successResponse(null);
+  }
+  validateCredentialRequest(): ErrorOrSuccess<null> {
+    return successResponse(null);
+  }
   getClientCredentialsById(): ErrorOrSuccess<IClientCredentials> {
     return errorResponse("No credentials found");
-  }
-  validate(): ErrorOrSuccess<null> {
-    return errorResponse("invalid credentials");
   }
 }
 
 class MockPassingClientCredentialsService implements IClientCredentialsService {
-  validate() {
+  validateTokenRequest() {
+    return successResponse(null);
+  }
+  validateCredentialRequest() {
     return successResponse(null);
   }
   getClientCredentialsById() {
--- a/src/functions/asyncCredential/asyncCredentialHandler.test.ts
+++ b/src/functions/asyncCredential/asyncCredentialHandler.test.ts
@@ -647,35 +647,35 @@ describe("Async Credential", () => {
     });
   });
 
-  // describe("Request body validation - using Client Credentials", () => {
-  //   describe("Given redirect_uri is present and does not match the registered redirect_uri value", () => {
-  //     it("Returns a 400 Bad Request response", async () => {
-  //       const event = buildRequest({
-  //         headers: { Authorization: `Bearer ${mockValidJwt}` },
-  //         body: JSON.stringify({
-  //           state: "mockState",
-  //           sub: "mockSub",
-  //           client_id: "mockClientId",
-  //           govuk_signin_journey_id: "mockGovukSigninJourneyId",
-  //           redirect_uri: "mockInvalidRedirectUri",
-  //         }),
-  //       });
-  //       dependencies.clientCredentialsService = () =>
-  //         new MockPassingClientCredentialsService();
+  describe("Request body validation - using Client Credentials", () => {
+    describe("Given redirect_uri is present and does not match the registered redirect_uri value", () => {
+      it("Returns a 400 Bad Request response", async () => {
+        const event = buildRequest({
+          headers: { Authorization: `Bearer ${mockValidJwt}` },
+          body: JSON.stringify({
+            state: "mockState",
+            sub: "mockSub",
+            client_id: "mockClientId",
+            govuk_signin_journey_id: "mockGovukSigninJourneyId",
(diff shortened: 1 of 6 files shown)


src/functions/asyncCredential/asyncCredentialHandler.test.ts
```diff
@@ -647,6 +647,36 @@ describe("Async Credential", () => {
     });
   });
 
+  // describe("Request body validation - using Client Credentials", () => {
```
sandy-may, line 650: Please may we not comment this test out - it's really important it constantly fails - both in the Github action and locally :)

J-son1: The pre-commit hook stops us from pushing with failing tests. Happy to comment out that part of the hook if needed

sandy-may: right you are - I assumed there was a reason!

src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -111,7 +112,37 @@ export async function lambdaHandler(
...
+  if (validateClientCredentialsResponse.value != null) {
+    return badRequestResponse({
+      error: "invalid_request",
+      errorDescription: validateClientCredentialsResponse.value,
+    });
+  }
+
+  let parsedRequestBody: IClientCredentials;
+  try {
+    parsedRequestBody = JSON.parse(requestBody);
```
sandy-may, line 130: The check to parse JSON should be much further up in this lambda. Please may we move it to the request body validation function if it's not there already,

src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -111,7 +112,37 @@ export async function lambdaHandler(
...
 
   const clientCredentials =
-    clientCredentialResponse.value as IClientCredentials;
+    clientCredentialResponse.value as IRegisteredClientCredentials;
+
+  const validateClientCredentialsResponse = clientCredentialsService.validate(
+    clientCredentials,
+    JSON.parse(requestBody),
+  );
+  if (validateClientCredentialsResponse.value != null) {
```
sandy-may, line 121: We should use the isError property on this, rather than the value property please

J-son1: Typescript still wants a null check on the value but I can add in the error property as well

src/functions/services/clientCredentialsService/clientCredentialsService.ts
```diff
@@ -60,6 +73,15 @@ export type IClientCredentials = {
   issuer: string;
   salt: string;
   hashed_client_secret: string;
+  redirect_uri?: string;
+};
+
+export type IRegisteredClientCredentials = {
```
sandy-may, line 79: I don't understand the need to have both of these types - please may we append the redirect_uri property to the IClientCredentials type?

The redirect_uri property is mandatory in SSM, but **not** mandatory in the request body to the credential lambda

src/functions/asyncCredential/asyncCredentialHandler.ts
sandy-may: We could cast this if we're confident in the value (which we should be as we control the types). It's how we've historically done this to retrieve the value.

src/functions/services/clientCredentialsService/clientCredentialsService.ts
J-son1: Updated to remove this interface.

src/functions/asyncCredential/asyncCredentialHandler.ts
J-son1: Updated to use `isError` property.

src/functions/asyncCredential/asyncCredentialHandler.ts
J-son1: Updated.

src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -67,8 +67,18 @@ export async function lambdaHandler(
     });
   }
 
+  let parsedRequestBody: ICredentialRequestBody;
+  try {
+    parsedRequestBody = JSON.parse(requestBody);
```
J-son1, line 72: May need a test for this

src/functions/services/clientCredentialsService/clientCredentialsService.test.ts
```diff
@@ -61,16 +72,46 @@ describe("Client Credentials Service", () => {
       });
     });
 
+    describe("redirect_uri validation", () => {
+      describe("Given redirect_uri is not present", () => {
+        it("Returns a log", () => {
+          mockStoredClientCredentials.redirect_uri = "";
```
J-son1, line 78: This can be removed now that the interface has been updated to have this field as an optional parameter

src/functions/services/clientCredentialsService/clientCredentialsService.test.ts
```diff
@@ -79,17 +120,64 @@ describe("Client Credentials Service", () => {
     });
   });
 
+  describe("Validate credential request credentials", () => {
+    describe("redirect_uri validation", () => {
+      describe("Given stored redirect_uri is not present", () => {
+        it("Returns a log", () => {
+          mockStoredClientCredentials.redirect_uri = "";
```
J-son1, line 127: This can be removed or maybe set to null or undefined now that the interface has been updated to have this field as an optional parameter

jamesamooney (approved):

PR #21, opened 2024-07-15
DCMAW-8264: Create session recovery service for credential lambda

jamesamooney:
### What changed
Create service to check if user already has a valid session, if so return success response with the `sub` from the `JAR` and "https://vocab.account.gov.uk/v1/credentialStatus": "pending"

### Why did it change
Following tech design which is found on ticket: DCMAW-8264

## Checklists

### DIFF
--- a/src/functions/asyncCredential/asyncCredentialHandler.test.ts
+++ b/src/functions/asyncCredential/asyncCredentialHandler.test.ts
@@ -55,7 +55,7 @@ const env = {
   ISSUER: "mockIssuer",
   SESSION_TABLE_NAME: "mockTableName",
   SESSION_TABLE_SUBJECT_IDENTIFIER_INDEX_NAME: "mockIndexName",
-  SESSION_RECOVERY_TIMEOUT: "mockSessionRecoveryTimeout",
+  SESSION_RECOVERY_TIMEOUT: "12345",
 };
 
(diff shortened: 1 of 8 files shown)


src/functions/asyncCredential/recoverSessionService/recoverSessionService.ts
```diff
@@ -0,0 +1,79 @@
...
+    let result: QueryCommandOutput;
+    try {
+      result = await dbClient.send(new QueryCommand(queryCommandInput));
+    } catch (error) {
+      return errorResponse(
+        "Unexpected error when querying session table whilst checking for recoverable session",
+      );
+    }
+
+    if (
```
jamesamooney, line 60: Change this check, was just initial testing out of stuff, not the best way to do this part

src/functions/asyncCredential/asyncCredentialHandler.test.ts
```diff
@@ -830,6 +912,97 @@ describe("Async Credential", () => {
       });
     });
   });
+
+  describe("Session recovery service", () => {
+    describe("Given service returns an unexpected error", () => {
```
jamesamooney, line 917: Given service returns error response or something similar

J-son1 (approved):

PR #24, opened 2024-07-17
DCMAW-8264: Add logs for missing environment variables in credential lambda

sandy-may:
DCMAW-8264
​
### What changed
Implements the logger for the async credential lambda
Adds logs for when there are missing/invalid environment variables
Moves mockLogger to its own file - prevents logger tests being automatically run when the file is imported

### Why did it change
Improve observability of failures and ensure tests correctly identify why an error occurs

## Checklists


src/functions/asyncCredential/asyncCredentialHandler.test.ts
```diff
@@ -174,6 +120,13 @@ describe("Async Credential", () => {
         const event = buildRequest();
         const result = await lambdaHandler(event, dependencies);
 
+        expect(mockLogger.getLogMessages()[0].logMessage.message).toBe(
+          "ENVIRONMENT_VARIABLE_MISSING",
```
sandy-may, line 124: mixed feelings about this - it is 'good enough' for me. Could update this to say ENVIRONMENT_VARIABLE_INVALID - it's a shared log message and keen not to split this validation out if possible.

jamesamooney: If you don't want to split the message and have one shared ENVIRONMENT_VARIABLE_INVALID message, we could also portentially log the value of the envVar, so when someone sees that's invalid log they can just look what the value was? Even this might be overkill as they'll already see what variable was invalid and can probs check and then work it out from there...

Will leave it up to you, I am happy with all options here

sandy-may: I'm not too keen on logging it out if possible unless we know we need to see it. We're the ones in control of it, so I'm happy enough with this approach

sandy-may: If it was an external team populating this value, I'd propose logging the value

src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -156,15 +151,15 @@ export async function lambdaHandler(
   }
 
   const recoverSessionService = dependencies.getRecoverSessionService(
-    sessionTableName,
-    sessionTableSubIndexName,
+    config.SESSION_TABLE_NAME,
```
jamesamooney, line 154: Is it worth deconstructing all these variables once after the config function? Suggestion only, feel free to dismiss

src/jest.config.ts
```diff
@@ -12,5 +12,6 @@ export default {
     ],
   ],
   testEnvironment: "node",
+  testMatch: ["**/*.test.ts"],
```
jamesamooney, line 15: Curious as to what prompted this?

sandy-may: Ah - was really curious why the logging.test.ts file was running for the lambda unit tests.

Thought I'd pin down our filters for running tests whilst I was debugging.

src/functions/asyncCredential/asyncCredentialHandler.ts
sandy-may: 🤷 if we want to. Will leave you to re-analyse this once you've finished all your dev. It adds extra lines of code

jamesamooney (approved):

PR #29, opened 2024-07-18
DCMAW-8265: Add test coverage for valid redirect_uri

sandy-may:
​DCMAW-8265
### What changed
Adds test coverage for when the supplied redirect_uri is registered in the ClientCredentials Array

### Why did it change
Flagged by Sonar

## Checklists

### DIFF
--- a/src/functions/asyncCredential/asyncCredentialHandler.ts
+++ b/src/functions/asyncCredential/asyncCredentialHandler.ts
@@ -138,7 +138,7 @@ export async function lambdaHandler(
 
   if (parsedRequestBody.redirect_uri) {
     const validateClientCredentialsResult =
-      clientCredentialsService.validateCredentialRequest(
+      clientCredentialsService.validateRedirectUri(
         clientCredentials,
         parsedRequestBody,
       );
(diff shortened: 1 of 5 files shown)


src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -138,7 +138,7 @@ export async function lambdaHandler(
 
   if (parsedRequestBody.redirect_uri) {
     const validateClientCredentialsResult =
-      clientCredentialsService.validateCredentialRequest(
+      clientCredentialsService.validateRedirectUri(
```
jamesamooney, line 141: This also validates the client secret... or are you thinking of this more like... the point is that we want to validate the redirect uri and the secret check is just to make sure we're looking at the right person's redirect uri?

sandy-may: The client id is validated here from what I can see: https://github.com/govuk-one-login/mobile-id-check-async/blob/1fa88cc1ca4b30f9542024bd1c22cffd3feb769e/src/functions/asyncCredential/asyncCredentialHandler.ts#L134

This lambda doesn't have access to the client secret - we're using the bearer token here rather than the client id/ client secret combination

jamesamooney: Sorry looking at completely the wrong function! As you were 😄

PR #33, opened 2024-07-23
DCMAW-8264: Prepare for monorepo

sandy-may:
​DCMAW-8264
### What changed

Move code into a backend-api folder 
Updates paths for:
- pre commit hook
- workflows
- sonar.properties

### Why did it change
Repository will become a monorepo with infrastructure code

## Checklists


.github/workflows/async-credential-push-to-main.yml
dothomson, line 1: Should the workflows be named after the stack they're deploying? 

Here I believe its inherited the repo name (sort of) but if we add in other stacks then it will be confusing.

jamesamooney:
As you've editted the pre-hook file, how would you feel about removing the npm run test from this? I don't find it helpful, especially when pairing. I think you should be able to push failing tests to a branch.

.github/workflows/async-credential-push-to-main.yml
sandy-may: yes

jamesamooney (approved):

PR #39, opened 2024-07-24
DCMAW-9802: Enforce Node 20

jamesamooney:
​DCMAW-9802

### Enforcing Node 20
- Added `engine` with `node` field to package.json
- Added `.npmrc` file

`engine` block and `npmrc` file enforces dev to use NodeJS `20.X.X`. To confirm, any minor or patch version of Node 20 will be allowed. Npm will not enforce it automatically with just the `engine` block, hence the need for an `.npmrc` file

### Easily install and use latest LTS Node 20 version
- Added `.nvmrc` file and specified `lts/iron` which is specifiying to look for the latest Node 20 LTS version.

Use commands `nvm install` and `nvm use` in `backend-api/src` to automatically install and use the latest node 20 version respectively. Node 20 is in LTS.

## Checklists


README.md
```diff
@@ -15,6 +15,7 @@ It follows the Client credentials grant flow from the OAuth2.0 Framework [see he
 - AWS SAM
 - node v.20
 - npm
+- nvm
```
dothomson, line 18: Nit: Technically not a dependency unless we're planning on managing multiple node versions in our scripts. It would be the developers responsibility to ensure they meet the `node v.20` [sic] requirement 

(Also other version managers are [available](https://github.com/Schniz/fnm))

README.md
```diff
@@ -15,6 +15,7 @@ It follows the Client credentials grant flow from the OAuth2.0 Framework [see he
 - AWS SAM
 - node v.20
```
dothomson, line 16:
```suggestion
- Node.js v20
```

dothomson (approved):

README.md
sandy-may: Agree with this - it is just for developer convenience if we use nvm (but good thought on even thinking about documentation).

What we **do** need is an engines property in our package.json to enforce the use of a certain node version (both locally and in the pipelines)

jamesamooney: Will remove this from the readme. @sandy-may re your point about the engines, did you see my notes on that? I think if we added the engines property there would have to be an instruction to devs to update their own config to force the correct node version

sandy-may: Shouldn't we always be forcing devs to user the correct node version?

README.md
jamesamooney: Done

sandy-may (approved):

PR #45, opened 2024-07-25
DCMAW-9802: Refactor ErrorOrValue pattern to improve typing

sandy-may:
​DCMAW-9802
### What changed
Introduce a Result type as a replacement for ErrorOrValue

### Why did it change

This is more type-safe and negates the need for casting

## Checklists


jamesamooney:
I general I like the naming.

`return error('foo');`

Does this make it seem like we're returning an actual `Error` object? The thought has likely come from the fact it used to be called `errorResponse`.

Not against this at all, just wanted thoughts.

sandy-may:
> In general I like the naming.
> 
> `return error('foo');`
> 
> Does this make it seem like we're returning an actual `Error` object? The thought has likely come from the fact it used to be called `errorResponse`.
> 
> Not against this at all, just wanted thoughts.

yep I'm thinking about errorResult()

backend-api/src/functions/types/result.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const success = <S>(value: S): SuccessResult<S> => {
+  return {
+    isError: false,
+    value,
+  };
+};
+
+export const error = <E>(value: E): ErrorResult<E> => {
```
J-son1, line 20: Should the <S> and <E> after the `=` signs be directly after the function name?

jamesamooney (approved):

PR #50, opened 2024-07-30
[DCMAW-9558] Deploy base stacks - dev

dothomson:
Ticket: https://govukverify.atlassian.net/browse/DCMAW-9558

This change introduces a terraform structure to assist with cloudformation deployments and to track the used parameters through github. 

The state for these terraform deployments is not intended to be stored between runs and instead will rely on terraform imports to recreate the state on each run.


infra/base_stacks/USAGE.md
sandy-may, line 1: 1) Is there a chance we could get a list of stacks that are deployed if we do the terraform apply?
2) Please may we get a terraform plan step in?

dothomson: > Is there a chance we could get a list of stacks that are deployed if we do the terraform apply? 

This is in the code and doesn't feel appropriate to put here in the generic usage section. 

> Please may we get a terraform plan step in?

I have added. The terraform apply does also show the plan and asks for input befor running.

sandy-may: We should have good documentation for what stacks this might deploy. I don't think we should look into code for this.

This readme **isn't** the right place for it - but we do need that documentation

infra/terraform/base_stacks/README.md
```diff
@@ -0,0 +1,34 @@
...
+- API Gateway logs
+- Certificate expiry
+- Grafana?
+- Checkov hook
+- lambda audit hook
+- infra audit hook
+
+- identity broke oidc provider ( different from github identity?)
+
+- ecr scan result logger?
```
sandy-may, line 19: this stack isn't great - it sends a very complex email that I find hard to read. I suggest we remove it if it has been deployed

infra/terraform/base_stacks/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Build Account Only:
+- GitHub Identity
+- Signer (also dev to allow deployments there?)
+- Container Signer
+
+Per Account:
+- API Gateway logs
+- Certificate expiry
+- Grafana?
```
sandy-may, line 12: This can go :) We don't use Grafana any more

infra/templates/testApp/template.yaml
```diff
@@ -0,0 +1,16 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
sandy-may, line 1: what is the test app for?

sandy-may (approved):

PR #62, opened 2024-08-05
[DCMAW-9808] Adding supporting infra

dothomson:
Ticket: https://govukverify.atlassian.net/browse/DCMAW-9808

Adding supporting infra for the backend api stack. This will include templates for the platform supporting stacks which aren't provided by devplatform

### DIFF
--- a/infra/templates/kms/template.yaml
+++ b/infra/templates/kms/template.yaml
@@ -1,0 +1,102 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Description: |
+  kms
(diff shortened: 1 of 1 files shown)


infra/templates/kms/template.yaml
```diff
@@ -0,0 +1,102 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Description: |
+  kms
```
Pranjal-GDS, line 4:
```suggestion
```

PR #66, opened 2024-08-06
DCAMW-9664: Add /async/credential contract tests for IPVCore consumer

jamesamooney:
​DCMAW-9664

### What changed

- Additions:
  - `/async/credential` endpoint to `createApp` function
  - Constructor function for dependency injection
  - State configuration class for test setup
- Extracts dependency mocks from handler tests to common location for use with contract tests
- Updates pact file

### Why did it change

This is part of the work to enable contract testing between IPVCore and ID Check.

## Checklists


backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialStateConfiguration.ts
```diff
@@ -0,0 +1,78 @@
...
+  SIGNING_KEY_ID: "mockKid",
+  ISSUER: "mockIssuer",
+  SESSION_TABLE_NAME: "mockTableName",
+  SESSION_TABLE_SUBJECT_IDENTIFIER_INDEX_NAME: "mockIndexName",
+  SESSION_TTL_IN_MILLISECONDS: "12345",
+  SQS_QUEUE: "mockSqsQueue",
+  CLIENT_REGISTRY_PARAMETER_NAME: "mockParmaterName",
+};
+
+export class AsyncCredentialStateConfiguration {
```
sandy-may, line 23: Let's chat about this - it doesn't have much to do with state and I'm not convinced this adds the right level of value

backend-api/src/functions/tests/pact/createApp.ts
```diff
@@ -12,7 +14,18 @@ export async function createApp(): Promise<Application> {
     const result = await asyncTokenHandlerConstructor({
       headers: req.headers,
       body: req.body,
-      dependencies: stateConfig.asyncTokenDependencies,
+      dependencies: asyncTokenStateConfig.dependencies,
+    });
+
+    res.status(result.statusCode);
+    res.send(JSON.parse(result.body));
```
sandy-may, line 21: What's the rationale for this?

sandy-may: I assume if anything we'd need to stringify it?

backend-api/src/functions/tests/pact/pact.test.ts
```diff
@@ -39,14 +40,22 @@ describe("Provider API contract verification", () => {
   it("validates adherence to all consumer contracts", () => {
     const stateHandlers = {
       "badDummySecret is not a valid basic auth secret": () => {
-        stateConfig.resetToPassingAsyncTokenDependencies();
-        stateConfig.asyncTokenDependenciesClientRegistryService = () =>
+        asyncTokenStateConfig.resetToPassingDependencies();
+        asyncTokenStateConfig.dependenciesClientRegistryService = () =>
```
sandy-may, line 44: This doesn't read very well for me - one to chat about :)

backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialStateConfiguration.ts
sandy-may, line 23: I think we could really simplify this -> one idea could be:

```
getDependenciesbyScenario(scenario: ScenarioType): IDependenices => {

switch(scenario) { 
   case 'INVALID_TOKEN': { 
      return {
tokenService: () => new MockInvalidTokenService
      break; 
   } 
   default: { 
      return {
tokenService: () => new MockTokenService()
      }
  }
}
   


```

sandy-may: ignoring formatting/syntax - but hopefully you get the gist

backend-api/src/functions/tests/pact/createApp.ts
J-son1: As an initial thought, it looks like it needs to be parsed so it can be asserted against in the test

backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialStateConfiguration.ts
J-son1: Renamed to remove 'state' for now

backend-api/src/functions/tests/pact/createApp.ts
J-son1: > As an initial thought, it looks like it needs to be parsed so it can be asserted against in the test

This is the case

backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialStateConfiguration.ts
J-son1: I've updated the function on this class to set the dependencies by scenario. Happy to discuss further if needed

J-son1: Update: Sonar doesn't like switch statements with fewer than 3 cases so I've changed the functions for to use if statements. Maybe something to review when more cases are added.

J-son1: Note: I believe state handlers are needed to setup the Lambdas to suit each interaction. This class is needed to ensure the setup of the Lambdas can be changed using state handlers in the pact.test.ts file.

backend-api/src/functions/tests/pact/asyncToken/asyncTokenConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+import { MockLoggingAdapter } from "../../../services/logging/tests/mockLogger";
+import { Logger } from "../../../services/logging/logger";
+import {
+  MockClientRegistryServiceBadRequestResult,
+  MockClientRegistryServiceSuccessResult,
+  MockTokenServiceSuccessResult,
+} from "../../../testUtils/asyncTokenMocks";
+import { IAsyncTokenRequestDependencies } from "../../../asyncToken/handlerDependencies";
+
+export class AsyncTokenConfiguration {
```
sandy-may, line 12: Please may we align this with our solution for the credential handler?

J-son1: I believe the solutions are currently the same. Is there anything in particular you want updated?

J-son1: Updated

backend-api/src/functions/testUtils/asyncCredentialMocks.ts
sandy-may, line 1: These aren't mocks for async credential - they're mocks for services that the async credential and async token at times use.

They're also not really a util - they're concrete implementations of a service. 

I'd suggest the following structure:

```
services
services/tests
services/tests/*.test.ts
service/tests/mocks.ts

```

sandy-may: This doesn't need to go in this PR

backend-api/src/functions/asyncCredential/handlerDependencies.ts
```diff
@@ -17,7 +17,7 @@ import {
 } from "./tokenService/tokenService";
 import { Logger } from "../services/logging/logger";
 
-export interface Dependencies {
+export interface IAsyncCredentialDependencies {
```
sandy-may, line 20: rational for the name change?

backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialConfiguration.ts
```diff
@@ -0,0 +1,68 @@
...
+  SIGNING_KEY_ID: "mockKid",
+  ISSUER: "mockIssuer",
+  SESSION_TABLE_NAME: "mockTableName",
+  SESSION_TABLE_SUBJECT_IDENTIFIER_INDEX_NAME: "mockIndexName",
+  SESSION_TTL_IN_MILLISECONDS: "12345",
+  SQS_QUEUE: "mockSqsQueue",
+  CLIENT_REGISTRY_PARAMETER_NAME: "mockParmaterName",
+};
+
+export class AsyncCredentialConfiguration {
```
sandy-may, line 23: Hello - I'm not tied to a particular pattern but I still think this is overly complicated and hard to work out what's going on

backend-api/src/functions/tests/pact/asyncCredential/asyncCredentialConfiguration.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor() {
+    this.dependencies = this.getPassingDependencies();
+  }
+
+  setDependenciesByScenario(scenario?: AsyncCredentialTestScenario) {
+    if (scenario === "badAccessToken is not a valid access token") {
+      this.resetToPassingDependencies();
+      this.dependencies.tokenService = () =>
+        new MockTokenServiceInvalidSignatureIPV();
```
sandy-may, line 35: I don't think we should be creating a mock for just IPVCore, otherwise it implies we're missing unit test coverage to me?

backend-api/src/functions/tests/pact/createApp.ts
```diff
@@ -12,7 +14,18 @@ export async function createApp(): Promise<Application> {
...
-      dependencies: stateConfig.asyncTokenDependencies,
+      dependencies: asyncTokenConfig.dependencies,
+    });
+
+    res.status(result.statusCode);
+    res.send(JSON.parse(result.body));
+  });
+
+  app.post("/async/credential", async (req: Request, res: Response) => {
+    const result = await asyncCredentialHandlerConstructor({
```
sandy-may, line 25: Do we feel there's value in abstracting the `asyncCredentialHandlerConstructor`? It feels a bit like it's creating extra code

sandy-may: removed when pairing

sandy-may (approved):

PR #68, opened 2024-08-07
[DCMAW-9786] Add dev platform assets to enable testing in codebuild

dothomson:
DCMAW-9786

This change includes building out the test container to be used in the dev, build and staging deployments. This is a pre-requisite for creating a post-merge Github action, which will be included in a subsequent PR.

### DIFF
--- a/backend-api/Dockerfile.test
+++ b/backend-api/Dockerfile.test
@@ -16,7 +16,7 @@ RUN aws --version
 
 # Copy the CloudFormation infrastructure template and application source code to the working directory
 COPY template.yaml /
-COPY src ./
+COPY template.yaml src openApiSpecs infra-tests eslint.config.mjs jest.config.ts tsconfig.json ./
 
 # Give user, 'test', permissions to execute test script and switch the user to 'test'
 COPY run-tests.sh /
--- a/backend-api/Dockerfile.test
+++ b/backend-api/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM node:iron-alpine AS builder
+FROM node:iron-alpine
 
 WORKDIR /backend-api
 
--- a/backend-api/Dockerfile.test
+++ b/backend-api/Dockerfile.test
@@ -7,7 +7,7 @@ RUN adduser --disabled-password test
 RUN chown test .
 
 COPY package.json package-lock.json ./
-RUN npm install
+RUN npm install --ignore-scripts
 
 
 ## Update container and install awscli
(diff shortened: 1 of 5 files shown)


backend-api/Dockerfile.test
```diff
@@ -0,0 +1,26 @@
+FROM node:iron-alpine AS builder
+
+WORKDIR /backend-api
+
+# Create a new user 'test' to avoid running the app as a root
+RUN adduser --disabled-password test
+RUN chown test .
+
+COPY package.json package-lock.json ./
+RUN npm install
```
sandy-may, line 10:
```suggestion
RUN npm install --ignore-scripts
```

sandy-may (commented):
This looks great. Please may we make sure this is covered by Sonar? The sonar properties file at the root of the repo has a sonar.sources property that wouldn't capture this directory.

I think it should to make sure we've caught any docker/.sh codesmells

backend-api/Dockerfile.test
```diff
@@ -0,0 +1,26 @@
+FROM node:iron-alpine AS builder
```
sandy-may, line 1: question - is `iron` general knowledge as a major version of Node? 

I'm wondering whether it'd be clearer to people looking at these files to reference the version number instead as the base image?

sandy-may: Separately - do we need a builder here - it looks unused

dtrts: Ah yes, i started with builder but it didn't seem to be doing anything which wouldn't be brought
over to the final image. I have remove that. 


Re: `iron`, I matched the value from `.nvmrc`. I would change both if i was to change it here.

sandy-may: `iron` - understood :) Happy to keep with what you've done

backend-api/run-tests-locally.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -eu
+
+stack_name=${1:-mob-async-backend}
+
+echo Running test against "${stack_name}"
```
dothomson, line 6: Nit: 
```suggestion
echo "Running test against ${stack_name}"
```

sandy-may: is that a self-nit?

dothomson: Always working on becoming a better dev.

sandy-may (approved):

PR #72, opened 2024-08-13
DCMAW-9911: Update status code when access token signature is invalid

sandy-may:
​DCMAW-9911
### What changed
Updates the response code in the /credential endpoint to a 400 when the access token signature is invalid

### Why did it change
Follows the pattern of being a user error. Decision record: https://gds.slack.com/archives/C066GM0H5GS/p1723197566263159

## Checklists


backend-api/src/functions/asyncToken/requestService/tests/mocks.ts
sandy-may, line 1: This helps remove the need for a Jest mock which I'd prefer not to use if poss

backend-api/package-lock.json
```diff
@@ -7,6 +7,7 @@
     "": {
       "name": "one-login-async-credentials",
       "version": "1.0.0",
+      "hasInstallScript": true,
```
sandy-may, line 10: Carry over from a previous change that was not committed in error

J-son1 (approved):

backend-api/src/functions/asyncCredential/requestService/requestService.ts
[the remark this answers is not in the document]
J-son1: Looks good to me! I'm happy with the refactor

PR #83, opened 2024-08-20
[DCMAW-9930] Set Up Platform DNS

mthkbarber:
### What changed
Set up infra for the new `review-b-async.account.gov.uk` DNS name.
Add `dns.tf` to the `base_stacks` module.
Create the `base_stacks_2` module for `dns_records.tf`
Add conditional to only use `*.review-b-async-${Environment}.account.gov.uk` in Dev
Add infra tests and combine the previously separated records and logging stacks into 1 template.

### Why did it change
For the new async backend journey.

Stacks deployed in Dev:

image

platform-dns infra tests:

image

platform-dns-records infra tests:

image

[DCMAW-9930](https://govukverify.atlassian.net/jira/software/c/projects/DCMAW/boards/438?label=Platform&selectedIssue=DCMAW-9930l)

[DCMAW-9930]: https://govukverify.atlassian.net/browse/DCMAW-9930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


infra/terraform/base_stacks/dns_logging.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_cloudformation_stack" "dns-logging" {
```
dothomson, line 1: A very minor thing: I generally follow the syntactic formatting of using underscores (`_`) in resource names. We then generally switch to hyphens for the name of the stack name as that is aws style. 



```suggestion
resource "aws_cloudformation_stack" "dns_logging" {
```

infra/terraform/base_stacks_2/dns_records.tf
```diff
@@ -0,0 +1,11 @@
+resource "aws_cloudformation_stack" "dns-records" {
+    name = "platform-dns-records"
+
+    template_body = file("../../templates/dns-records/template.yaml")
+
+    parameters = {
+        Environment = var.environment
+    }
+
+    depends_on = [aws_cloudformation_stack.dns-records]
```
dothomson, line 10: This "depends_on" is not required as it is only referencing itself. It might not even pass `terraform validation` 

Terraform's job is to find the dependencies between different resources. Typically this is done implicitly if a resource depends on the output of another (for example passing a security group name to a lambda vpc config, terraform will make the security group first) 

Sometime you require resources to be made in a particular order without them having an explicit reference between the two. In cases like this you would add the depends on to enforce that relationship. 

An example might be two cloud fromation stacks, with one using an `!ImportValue`  of an output from another. There is no relationship between the two as defined in terraform but there is a relationship between the two in AWS. Here you would use the depends_on clause to ensure they are deployed correctly. 

Since this terraform module has a single resource to deploy, the depends_on has no functionality. The dependency tree here is managed by the fact we won't deploy `base_stacks_2` until we have completed our dns record deployment into the dev platform account managing the route 53 record.

dothomson, line 10:
```suggestion
```

infra/terraform/base_stacks_2/main.tf
dothomson, line 1: Lines 2-5 are not used. I would consider removing them.

infra/terraform/base_stacks/dns_logging.tf
dothomson, line 1: Nit: There's no real need for separate files for these two stacks. I would group them together personally.

infra/templates/dns/template.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+Resources:
+  PublicHostedZone:
+    Type: AWS::Route53::HostedZone
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
+    Properties:
+      Name: !If
+        - IsProduction
+        - review-b-async.account.gov.uk
```
dothomson, line 27: Can we check with @sandy-may  @Pranjal-GDS about naming? 
I've now started to see `id-check-v2` being used alongside async so i'm worried about more fractured naming.

mthkbarber: This is the name I was instructed to use by @sandy-may

dothomson: Sounds good. I'm sure it is correct for the domains, but as i was introduced to this new name recently I'm also keen to flag it to the wider team about what names are in use for our team and product.

infra/terraform/base_stacks/dns_logging.tf
mthkbarber: Done.

infra/terraform/base_stacks_2/dns_records.tf
mthkbarber: Removed.

infra/terraform/base_stacks_2/main.tf
mthkbarber: Removed.

infra/templates/dns/template.yaml
sandymay-dd: I'm double checking on our product name in ID Check

ID Check V2 is the new production/app name for the entirety of mobile.

infra/templates/dns-records/template.yaml
```diff
@@ -0,0 +1,100 @@
...
+      DomainName: !Sub
+      - "${DNSSUFFIX}"
+      - DNSSUFFIX: !FindInMap [ PlatformConfiguration, !Ref Environment, DNSSUFFIX ]
+
+      SubjectAlternativeNames:
+      - !Sub
+        - "api.${DNSSUFFIX}"
+        - DNSSUFFIX: !FindInMap [ PlatformConfiguration, !Ref Environment, DNSSUFFIX ]
+      - !Sub
+        - "www.${DNSSUFFIX}"
```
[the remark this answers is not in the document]
sandymay-dd: No frontend needed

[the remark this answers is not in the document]
sandymay-dd: `Can we verify which domains we want to use?`

I'm not sure what this means I'm afraid

infra/templates/dns/template.yaml
sandymay-dd: ID check V2 has gone as a product name - I take it all back.

id-check-async is the way to go!

infra/terraform/base_stacks_2/README.md
```diff
@@ -0,0 +1,40 @@
+# Before applying changes
+
+For initial configuration please refer to the README file located in ./infra/terraform
+
+This base_stacks_2 directory is only to be deployed after the deployment of the original base_stacks as dns_records.tf is dependent on dns.tf being deployed first.
+
+# Applying changes
```
dothomson, line 7: Please remove these instructions as it is duplicating the details from the `./terraform/README.md` (née `USAGE.md`) document. The aim is to keep both the `base-stacks` and `base-stacks-2` terraform configurations structured in the same way so we should be able to cover both with the same document. 

---- 

My preference would be to keep the `USAGE.md` document as a concise reference for the commands, with a `README.md` to detail out more the reasoning for each tf-config, but I'm also happy to follow your lead with any improvements to the documentation.

infra/templates/dns-records/template.yaml
```diff
@@ -0,0 +1,94 @@
...
+        Product: "ID Check Async"
+        System: "ID Check Async"
+        Environment: !Sub "${Environment}"
+
+Outputs:
+  CertificateARN:
+    Description: The ARN for the ACM Certificate created in this template
+    Value: !Ref ExternalCertificate
+    Export:
+      Name: platform-dns-records-certificate-arn
```
dothomson, line 94:
```suggestion
      Name: !Sub ${AWS:StackName}-certificate-arn
```

dothomson: (Assuming that syntax is correct.)

infra/terraform/base_stacks_2/README.md
mthkbarber: Yep makes sense!

I only renamed the `USAGE.md` to `README` as its a mono repo and the use of `README` already seems established, happy to revert it back to `USAGE` if we also potentially end up adding a `README`.

infra/templates/dns-records/template.yaml
```diff
@@ -0,0 +1,94 @@
...
+            - "${DNSSUFFIX}"
+            - DNSSUFFIX: !FindInMap [ PlatformConfiguration, !Ref Environment, DNSSUFFIX ]
+        - Key: Product
+          Value: "ID Check Async"
+        - Key: System
+          Value: "ID Check Async"
+        - Key: Environment
+          Value: !Sub "${Environment}"
+
+  CertificateARNSSM:
```
dothomson, line 76: I'm tempted to suggest we remove this parameter. 

I've had a brief discussion with @sandymay-dd about how and when to use parameters verses exports verses hardcoding values. We never settled on any guidelines so I welcome your input on what you'd like to do in this case.

Here I'm considering how we might want to access the certificate arn. 

When using it in another cloudformation stack then using the export prevents us from deleting the cloudformation stack until all usages of it is gone. I don't think happens with a parameter. 

On the other hand a parameter is valuable when retrieving the value outside of cloudformation, so potentially used when configuring a containers environment variables for example, or accessing the value as part of a lambda run.

dothomson: Actually, ignore all of the above. Looking through the existing code base the certificate is only accessed through the parameter. I trust they gave this decision more thought than me so I think we should stick to that approach.

infra/terraform/base_stacks_2/README.md
dothomson: As long as the document name and its contents / intention line up then we're in a good place.

infra/templates/dns-records/template.yaml
mthkbarber: Yep! Amended!

infra/terraform/base_stacks_2/README.md
```diff
@@ -0,0 +1,5 @@
+# Before applying changes
+
+For initial configuration please refer to the README file located in ./infra/terraform
+
+This base_stacks_2 directory is only to be deployed after the deployment of the original base_stacks as dns_records.tf is dependent on dns.tf being deployed first.
```
dothomson, line 5: Please add in details about the dependency on the configuration in the devplatform account which hosts the base route53 record. 

Having a dependency on the `dns` stack is not sufficient to explain why this is its own terraform configuration.

infra/terraform/base_stacks_2/main.tf
```diff
@@ -0,0 +1,22 @@
...
+
+provider "aws" {
+  region = "eu-west-2"
+
+  allowed_account_ids = [local.account_vars[var.environment].account_id]
+}
+
+terraform {
+  backend "s3" {
+    # bucket = "dmica-${local.environment}-tfs" # set by ../_backend/${local.environment}.s3.tfbackend as variables are unable to be used here
```
dothomson, line 19: This is referencing some potential bucket names form my spike but is likely more confusing now. Please change this to something clearer. 
Potentially: 
```suggestion
    # "bucket" name provided at run time. Check the readme for details.  
```

or something similar.

infra/templates/dns/template.yaml
dothomson, line 1: Nit: Have you used the `rain` tool before? 
https://github.com/aws-cloudformation/rain

I'd like to use the `rain fmt` command more regularly going forward. Lets give these a `fmt` before merging.

mthkbarber: Done.

infra/terraform/base_stacks_2/main.tf
mthkbarber: Done.

dothomson (approved):

PR #86, opened 2024-08-21
DCMAW-9786: Update the backend-api token endpoint so it can be deployed locally

sandymay-dd:
​DCMAW-9786

### What changed
- Adds missing infrastructure components (SQS/DLQ/IAM policies/Environment variables
- Fixes order of arguments passed into the lambda handler constructor
- Migrate to use Secrets Manager rather than SSM

### Why did it change
So that the application can be deployed
### Evidence
Token lambda can be triggered in the ui
image

## Checklists


backend-api/template.yaml
```diff
@@ -225,4 +270,77 @@ Resources:
...
+              AWS:
+                - !FindInMap [TxMA, !Ref Environment, TxmaAccount]
+
+  TxMASQSQueueDeadLetterQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      MessageRetentionPeriod: 259200 # three days
+      KmsMasterKeyId: !Sub TxMAKMSEncryptionKey
+
+  TxMAKMSEncryptionKey:
```
sandymay-dd, line 314: I _think this should live in the KMS stack, but it means we'll need to create a new key. It doesn't in the other backend repository.

@dothomson ?

backend-api/template.yaml
```diff
@@ -225,4 +270,77 @@ Resources:
...
+              - "kms:Encrypt"
+              - "kms:Decrypt"
+              - "kms:GenerateDataKey"
+            Resource:
+              - "*"
+      KeySpec: SYMMETRIC_DEFAULT
+      KeyUsage: ENCRYPT_DECRYPT
+      MultiRegion: false
+      PendingWindowInDays: 7
+      Tags:
```
sandymay-dd, line 342: Is there a tagging strategy?

backend-api/template.yaml
dothomson: I must have made loads of those stacks and the deletion window hasn't caused any issues. My current approach to KMS keys definitely isn't consistent. 

I don't see anything hugely wrong with having it here.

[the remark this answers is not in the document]
dothomson: Yes please. They provide a slightly nice UI experience also.

backend-api/template.yaml
```diff
@@ -18,6 +18,18 @@ Mappings:
...
+      TxmaAccount: arn:aws:iam::750703655225:root
+    staging:
+      TxmaAccount: arn:aws:iam::178023842775:root
+    integration:
+      TxmaAccount: arn:aws:iam::729485541398:root
+    production:
+      TxmaAccount: arn:aws:iam::451773080033:root
+
 Parameters:
   Environment:
```
dothomson, line 34: Can we get some `AllowedValues` on the environment to limit it to our environment names. especially if they're being used to pull out variables from the mappings.

backend-api/template.yaml
```diff
@@ -225,4 +270,77 @@ Resources:
...
+      MessageRetentionPeriod: 604800
+      VisibilityTimeout: 60
+      KmsMasterKeyId: !Ref TxMAKMSEncryptionKey
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt TxMASQSQueueDeadLetterQueue.Arn
+        maxReceiveCount: 5
+
+
+# TxMA account consumes this queue
+  TxMASQSQueuePolicy:
```
sandymay-dd, line 288: I don't think we should be giving TxMA access to Dev/Build so I'll move this into a conditional

backend-api/template.yaml
dothomson: I currently don't have one. I'm not sure if there's a programme wide one. It would be nice to look up some best practices and pick what works for us.

backend-api/template.yaml
sandy-may: added for TxMA

backend-api/template.yaml
sandy-may: added

backend-api/template.yaml
sandy-may: https://gds-way.digital.cabinet-office.gov.uk/manuals/aws-tagging.html#tagging-aws-resources

backend-api/template.yaml
sandymay-dd: added for all - just testing!

backend-api/template.yaml
sandymay-dd: @dothomson -> I think the sam app inherits tags from the pipeline. the stack orchestration tool uses the` aws cloudformation create-stack` command, which takes in a `--tags` argument.

I _assume_ the resources in the sam app also inherit the tags if the resources are tag-able

Potentially something we should be doing in tf instead of the template.yaml

backend-api/template.yaml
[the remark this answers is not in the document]
sandymay-dd: done

backend-api/template.yaml
sandymay-dd: I've taken tags out as I think these should be provided at a stack level and inherited from the pipelines

backend-api/template.yaml
sandymay-dd: updated

backend-api/template.yaml
dothomson, line 1: Nit: Have you tried using `rain fmt` on this template? 
https://aws-cloudformation.github.io/rain/rain_fmt.html

I know mobile platform use `rain` to help manage their templates (format and splitting i think).

sandymay-dd: nada - i can

dothomson: Here is a PR showing the effects of it: https://github.com/govuk-one-login/mobile-id-check-async/pull/87 
I see that it is quite opinionated in its choices which I always like. 

Nothing important for this PR but a nice tool to look at.

backend-api/template.yaml
```diff
@@ -216,13 +294,134 @@ Resources:
...
-        - PolicyName: AsyncTokenFunctionKmsPolicy
-          PolicyDocument:
-            Version: '2012-10-17'
-            Statement:
-              - Effect: Allow
-                Action:
-                  - kms:Sign
-                Resource: !ImportValue platform-kms-signing-key-arn
\ No newline at end of file
+                  - !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${Environment}/clientRegistry*"
```
dothomson, line 297: Why the wildcard at the end? Are we expecting multiple secrets with the same prefix?

sandymay-dd: AWS adds a random 6 character string to the end of secrets - do you know if there's a way round this?

backend-api/template.yaml
sandymay-dd: I like opinionated formats. Let me install rain

backend-api/template.yaml
dothomson: Ah I follow now. 

I checked their docs for guidance and it is the same process, they just has a slightly more constrained approach: 
https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_examples.html#auth-and-access_examples_wildcard

image

So this best we could get is: 

```suggestion
                  - !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${Environment}/clientRegistry-??????"
```

Since AWS always adds those 6 characters this blocks anyone sneaking a different secret in.

dothomson (approved):

backend-api/template.yaml
```diff
@@ -216,13 +294,134 @@ Resources:
...
+            Principal:
+              AWS: !Sub arn:aws:iam::${AWS::AccountId}:root
+            Action:
+              - kms:*
+            Resource:
+              - '*'
+      KeySpec: ECC_NIST_P256
+      KeyUsage: SIGN_VERIFY
+      MultiRegion: false
+      PendingWindowInDays: 7
```
dothomson, line 417: I feel like 7 days is very small for production keys. Whilst I'm sure we would realize something has gone wrong in the 7 days lets increase this to the maximum for our prod accounts (or all non-dev?)

dothomson (approved):

backend-api/template.yaml
sandymay-dd: kept as 7 in dev

sandymay-dd: raised to 30 in all other envs

backend-api/template.yaml
sandymay-dd: done!

PR #106, opened 2024-09-03
DCMAW-9786: Upload test image and sam artifact to build

sandymay-dd:
​DCMAW-XXXXX

## Checklists

### DIFF
--- a/.github/workflows/backend-api-push-to-main.yml
+++ b/.github/workflows/backend-api-push-to-main.yml
@@ -54,7 +54,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BACKEND_API_DEV_TEST_IMAGE_REPOSITORY: ${{ secrets.BACKEND_API_DEV_TEST_IMAGE_REPOSITORY }}
-      BACKEND_API_DEV_COSIGN_KEY: ${{secrets.BACKEND_API_DEV_COSIGN_KEY}}
+      BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY: ${{ secrets.BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY }}
+      ASYNC_BUILD_COSIGN_KEY: ${{secrets.ASYNC_BUILD_COSIGN_KEY}}
       IMAGE_TAG: latest
     defaults:
       run:
@@ -69,7 +70,7 @@ jobs:
 
     - name: Build test image
       run: |
-        docker build -t $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG . 
+        docker build -t $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG -t $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG . 
 
     - name: Configure AWS credentials for DEV
       uses: aws-actions/configure-aws-credentials@55f725fcb586ca16a5ed5b0d75d464defbfa831b #main
@@ -89,10 +90,13 @@ jobs:
       with:
         cosign-release: 'v1.9.0'
 
-    # Commenting out due to AWS permission issues
-    # - name: Sign the image in DEV
-    #   run: |
-    #     cosign sign --key awskms:///$BACKEND_API_DEV_COSIGN_KEY $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
+    - name: Sign the image in BUILD
+      run: |
+        cosign sign --key awskms:///$ASYNC_BUILD_COSIGN_KEY $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
+
+    - name: Push image to BUILD
+      run: |
+        docker push $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
 
   build-and-upload-sam-artifact:
     name: Validate & upload S3 artifact to dev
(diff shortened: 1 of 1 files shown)


dothomson (commented):
Does the `workflow_dispatch` allow you to run this workflow from a branch? 
If so there might be a small risk of triggering a prod deployment from a branch. I know there's other limitations elsewhere but would be good to verify.

sandymay-dd:
> Does the `workflow_dispatch` allow you to run this workflow from a branch? If so there might be a small risk of triggering a prod deployment from a branch. I know there's other limitations elsewhere but would be good to verify.

Yes it would allow it - I've tested it. I'm happy to remove it

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -54,7 +54,8 @@ jobs:
     runs-on: ubuntu-latest
```
mthkbarber, line 54: I think these Dev and Build `push-and-test-image` should be in separate jobs. I can see the the Dev `BACKEND_API_DEV_COSIGN_KEY` has been deleted which would be needed for to sign the image in Dev, though this is currently commented out

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -69,7 +70,7 @@ jobs:
 
     - name: Build test image
       run: |
-        docker build -t $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG . 
+        docker build -t $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG -t $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG .
```
mthkbarber, line 73: Are there any implications of us pushing a Docker Image to Build with a Dev tag and vice versa?

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -144,3 +148,17 @@ jobs:
           signing-profile-name: ${{ secrets.BACKEND_API_DEV_SIGNING_PROFILE }}
           template-file: .aws-sam/build/template.yaml
           working-directory: backend-api
+
+      - name: Configure AWS credentials
```
mthkbarber, line 152: Should we move these into a separate job `build-and-upload-sam-artifact-to-build`?

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -144,3 +148,17 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@55f725fcb586ca16a5ed5b0d75d464defbfa831b #main
+        with:
+          aws-region: eu-west-2
+          role-to-assume: ${{ secrets.BACKEND_API_BUILD_GH_ACTIONS_ROLE_ARN }}
+
+      - name: Upload SAM artifact into the BUILD artifact bucket
+        uses: govuk-one-login/devplatform-upload-action@6985ccbaa306e2320a8826252c922af65242b283 #main
+        with:
+          artifact-bucket-name: ${{ secrets.BACKEND_API_BUILD_ARTIFACT_BUCKET }}
+          signing-profile-name: ${{ secrets.BACKEND_API_BUILD_SIGNING_PROFILE }}
```
mthkbarber, line 162:
```suggestion
          signing-profile-name: ${{ secrets.BUILD_SIGNING_PROFILE_NAME }}
```

With the secret amendment

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -144,3 +148,17 @@ jobs:
           signing-profile-name: ${{ secrets.BACKEND_API_DEV_SIGNING_PROFILE }}
```
mthkbarber, line 148:
```suggestion
          signing-profile-name: ${{ secrets.DEV_SIGNING_PROFILE_NAME }}
```

With the secrets name changes.

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -54,7 +54,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BACKEND_API_DEV_TEST_IMAGE_REPOSITORY: ${{ secrets.BACKEND_API_DEV_TEST_IMAGE_REPOSITORY }}
-      BACKEND_API_DEV_COSIGN_KEY: ${{secrets.BACKEND_API_DEV_COSIGN_KEY}}
+      BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY: ${{ secrets.BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY }}
+      ASYNC_BUILD_COSIGN_KEY: ${{secrets.ASYNC_BUILD_COSIGN_KEY}}
```
mthkbarber, line 58:
```suggestion
      BUILD_CONTAINER_SIGN_KMS_KEY: ${{secrets.BUILD_CONTAINER_SIGN_KMS_KEY}}
```

Updated secret name

.github/workflows/backend-api-push-to-main.yml
```diff
@@ -89,10 +90,13 @@ jobs:
       with:
         cosign-release: 'v1.9.0'
 
-    # Commenting out due to AWS permission issues
-    # - name: Sign the image in DEV
-    #   run: |
-    #     cosign sign --key awskms:///$BACKEND_API_DEV_COSIGN_KEY $BACKEND_API_DEV_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
+    - name: Sign the image in BUILD
+      run: |
+        cosign sign --key awskms:///$ASYNC_BUILD_COSIGN_KEY $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
```
mthkbarber, line 95:
```suggestion
        cosign sign --key awskms:///$BUILD_CONTAINER_SIGN_KMS_KEY $BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY:$IMAGE_TAG
```

.github/workflows/backend-api-push-to-main.yml
sandymay-dd: Happy to do this. I did this to reduce the workflow time and avoid caching the docker image + sam application (which is possible, but adds a bit more complexity)

The reason I think it would be good to separate it would be because we could run the jobs in parallel and they shouldn't depend on each other given Dev is not part of our route to live.

What's your rationale for splitting it out, out of curiosity?

.github/workflows/backend-api-push-to-main.yml
sandymay-dd: I think it's only adding the `latest` tag here (?)

mthkbarber: Sorry I should have been clearer in my comment it relates more to the name of the image `$BACKEND_API_DEV_TEST_IMAGE_REPOSITORY` and `$BACKEND_API_BUILD_TEST_IMAGE_REPOSITORY` as opposed to `latest`.

I can't see anything in ECR that would but just curious

.github/workflows/backend-api-push-to-main.yml
mthkbarber: Because it is an already established pattern within Mobile Pod. 😅 

It will help us identify when and where this GitHub pipeline breaks more easily and only re-run jobs that have failures.

I am unsure if there are any form of implications (security wise) from using both Dev and Build environment variables in the same job but seems cleaner to me to separate them out.

Might be a better question for @Pranjal-GDS as to why we have chosen this method?

mthkbarber:
@sandymay-dd looks good to me, I may be misremembering but I thought we agreed to do the Dev and Build ECR jobs in parallel and then do the same for pushing the SAM artifacts to Dev and Build?

sandymay-dd:
> @sandymay-dd looks good to me, I may be misremembering but I thought we agreed to do the Dev and Build ECR jobs in parallel and then do the same for pushing the SAM artifacts to Dev and Build?

I thought it was the opposite as we were keen on splitting out Dev and Build!!

mthkbarber (approved):

PR #165, opened 2024-09-22
DCMAW-8965: issue a service token (part 2)

ahickmann-gds:
​DCMAW-8965

### What changed
**JWE**
- Fetch async backend public encryption key from JWKS endpoint
- Generate a JWE 

**Other changes**
- Rename `MOCK_STS_BASE_URL` to `STS_MOCK_BASE_URL`
- Add API tests 
- Add STS mock API DNS
- Move `infrastructure-tests` to `tests` folder
- Validate request body params `grant_type` and `subject_token_type`

Feature deployed to `dev` and tested by sending a valid `POST` request to `/token`:
Screenshot 2024-09-23 at 18 45 07

## Checklists


sts-mock/template.yaml
```diff
@@ -147,10 +196,30 @@ Resources:
       Role: !GetAtt TokenFunctionLambdaRole.Arn
       Environment:
         Variables:
-          # MOCK_STS_BASE_URL to be updated once available
-          MOCK_STS_BASE_URL: MockStsBaseUrl
-          ASYNC_BACKEND_BASE_URL: !Sub https://mob-async-backend.review-b-async.${Environment}.account.gov.uk
+          STS_MOCK_BASE_URL: !Sub https://${StsMockApiDomainName}
+#          ASYNC_BACKEND_BASE_URL: !Sub https://public-mob-async-backend.review-b-async.${Environment}.account.gov.uk
```
ahickmann-gds, line 200: Can we please have a discussion about this value? cc @sandymay-dd

sts-mock/src/token/tokenEncrypter/tokenEncrypter.ts
```diff
@@ -0,0 +1,106 @@
...
+    } catch {
+      return errorResult({
+        errorMessage: "Unexpected network error fetching JWKS",
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+
+    let body: object;
+    try {
+      body = JSON.parse(await response.text());
```
jmooney-dd, line 66:
```suggestion
      body = await response.json();
```

jmooney-dd: We found this on Friday when we were making a similar func. Not better just thought I'd share it, both ways work just fine 😄

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
```diff
@@ -0,0 +1,220 @@
+import { TokenEncrypter } from "../tokenEncrypter";
+
+describe("Token Encrypter", () => {
+  let tokenEncrypter: TokenEncrypter;
+
+  const dummyJwt = "header.payload.signature";
+
+  beforeEach(() => {
+    tokenEncrypter = new TokenEncrypter("dummyJwksUrl");
```
jmooney-dd, line 9: Small nit but would you prefer the dummyJwksUrl to be in the shape of a URL as that's what it is meant to be? e.g. https://dummyJwksUrl.com? Appreciate it doesn't affect the behavior of the test though so happy for you to resolve if you don't see enough value

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+      expect(result.isError).toBe(true);
+      expect(result.value).toStrictEqual({
+        errorMessage: "No encryption key in JWKS",
+        errorCategory: "SERVER_ERROR",
+      });
+    });
+  });
+
+  describe("Given the encryption JWK is invalid and therefore cannot be parsed as a KeyObject", () => {
```
jmooney-dd, line 142: What is the difference between this test and "Given the encryption JWK is invalid as it's missing required fields" (test below)

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+      expect(result.isError).toBe(true);
+      expect(result.value).toStrictEqual({
+        errorMessage: "Error creating public encryption key",
+        errorCategory: "SERVER_ERROR",
+      });
+    });
+  });
+
+  describe("Given an error happens trying to encrypt the JWT (e.g. JWK type is not compatible with encryption algorithm)", () => {
```
jmooney-dd, line 182: It says "e.g. JWK type is not compatible with encryption algorithm"

Is this the specific reason THIS test is failing? If it is, are there other scenarios we should test specifically?

If not would it be better worded as Given there is an unexpected error trying to encrypt etc.. 

The wording of eg is just throwing me a little

sts-mock/src/token/configService/configService.ts
```diff
@@ -12,9 +12,9 @@ export interface IGetConfig<T> {
 
 export class ConfigService implements IGetConfig<Config> {
```
jmooney-dd, line 13: Once you've checked if the URL envVars are there, you could also do something like:

```
try {
      new URL(env.STS_MOCK_BASE_URL);
    } catch {
      return errorResult({
        errorMessage: "STS_JWKS_ENDPOINT is not a URL",
      });
    }
```

jmooney-dd: That way if it's not a valid URL we're not bothering executing the rest of the lambda since it won't work anyway

sts-mock/src/token/tokenEncrypter/tokenEncrypter.ts
ahickmann-gds: Yes, I did use `response.json()` initially, but I switched back to `JSON.parse()` because I am more familiar with it, but I can change it.

ahickmann-gds: Updated.

sts-mock/template.yaml
ahickmann-gds: Dealt with.

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
ahickmann-gds: None - thanks for spotting that. I have removed one of them.

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
jmooney-dd: Although might affect the behavior if you do change your config service to check if it's a valid URL

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
ahickmann-gds: I hear you. But we don't need to test other behaviours because we would be testing the library (`jose`) and hopefully their developers have done that. 😅 Ideally I would mock `jose` to throw an error (so that the error really looks like something unexpected has happened) but because we don't particularly like using `jest.mocks` I thought I'd just cause the error myself, and I did that by making the key type not compatible with the encryption algorithm because I knew the library would "complain". I thought I'd include this in the test title to make it clear, but perhaps a comment is better?

This error would not happen in reality unless we modify the code.

sts-mock/src/token/configService/configService.ts
ahickmann-gds: We could add this additional check but I don't think it's worth the time considering that this is a mock that will be used by developers. Those env var values are defined in the SAM template and we shouldn't be "messing" with them. I think there are so many things we could check but sometimes checking for everything doesn't save time. What do you think? 🤔

sts-mock/template.yaml
```diff
@@ -291,4 +329,12 @@ Resources:
...
               StringEquals:
-                aws:SourceAccount: !Ref AWS::AccountId
\ No newline at end of file
+                aws:SourceAccount: !Ref AWS::AccountId
+
+Outputs:
+  StsMockApiUrl:
+    Description: STS Mock API Gateway DNS
+    Value: !Sub https://${StsMockApiDomainName}
+    Export:
```
ahickmann-gds, line 338: I thought it might be nice exporting this so that it can be referenced in the async backend SAM template - but I can remove it if not necessary/won't work...

sandymay-dd: We should export it, at least so someone can easily see in Cloudformation what the DNS is. We need to be wary about importing across templates as it creates tight application couplings, despite its benefits. One to figure out as we go along

sandymay-dd: What we will definitely want is a script that can:
1) Infer what the DNS should be based off the format we've used
2) Deploy the application stacks in parallel

ahickmann-gds: Good point 🤔 -  in that case just ignore me 😄

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
sandymay-dd:
```suggestion
// This tests what happens in our codebase if jose throws an error. It does not test every scenario for which jose might throw an error as this would be testing the underlying functionality of jose. The example below is a scenario where the JWK type is not compatible with the encryption algorithm, which is one scenario for which jose might throw an error.

describe("Given there is an unexpected error encrypting the JWT", () => {
```

@jmooney-dd - I think this is what we're trying to convey here - very keen we don't start testing jose. Does this help?

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
sandymay-dd: We've been bitten (gently) once before with this when we thought we had regression coverage using our unit tests but our test data wasn't in the right format. Given it's hopefully very low effort - please may we structure this as a URL? I appreciate this doesn't change the behaviour of our tests/codebase at all though!

sts-mock/src/token/configService/configService.ts
sandymay-dd: I think let's just get this in as-is - it's a mock asset and if we've got an incorrect URL the request to the well-known endpoint will fail with an error. If it becomes a regular issue (e.g. if it's very easy to make mistakes when deploying dev stacks) let's add it.

sandymay-dd (dismissed):
Nice work

sts-mock/src/token/tokenEncrypter/tests/tokenEncrypter.test.ts
jmooney-dd: Yeah this all makes sense, thanks both for explaining!

ahickmann-gds: Test description updated and a comment added.

sandymay-dd (approved):

PR #167, opened 2024-09-23
DCMAW-10243: Simplifying DNS Cloudformation outputs

sandymay-dd:
​DCMAW-10243

### What changed
Simplifying the output for pretty DNS in the backend-api template.yaml following a lovely suggestion from @ahickmann-gds 

### Why did it change
Avoids repeating code and locking into the implementation

### Testing
image
image

## Checklists


ahickmann-gds (approved):

ahickmann-gds:
Thank you! 😍

backend-api/template.yaml
```diff
@@ -1158,18 +1158,9 @@ Resources:
 Outputs:
   SessionsApiUrl:
```
ahickmann-gds, line 1159: I've approved this PR but I thought just now that it might be nice to export this value so that I can reference it in the `sts-mock` template like this `ASYNC_BACKEND_BASE_URL: !ImportValue mob-async-backend-SessionsApiUrl`

What do you think? 🤔 I think this should work...

```
    Export:
      Name:
        Fn::Sub: ${AWS::StackName}-SessionsApiUrl
```

sandymay-dd: Left a similar comment on your PR - we need to avoid tight coupling of stacks otherwise you enforce the deployment order of the stacks (i.e. deploy the sts-mock stack first and then deploy the backend-api stack)

sandymay-dd: Will avoid for now - unless we find it becomes needed :)

ahickmann-gds: Sounds good to me :)

PR #171, opened 2024-09-24
[DCMAW-8233] Create high and low threshold alarms for jwks endpoint

mthkbarber:
​DCMAW-8233

### What changed
Create high and low threshold alarms for the `/.well-known/jwks.json` endpoint.

Alarms will trigger if we either have:
80% or more of invocations return a 5XX error.

We have 2 or more 5XX errors within 5 minutes in 2 periods.

Alarms are both currently set to show up in warning channels, High threshold alarm will need to be moved to the critical SNS topic at some point

### Why did it change
To notify us when services go down.

image

Test passing when runbook set to true and warning alarm name is used:

image

Test fails when runbook set to false and warning alarm name is used:

image


backend-api/template.yaml
```diff
@@ -1112,6 +1112,126 @@ Resources:
...
+          ReturnData: false
+          MetricStat:
+            Metric:
+              Namespace: AWS/ApiGateway
+              MetricName: Count
+              Dimensions:
+                - Name: Method
+                  Value: GET
+                - Name: ApiName
+                  Value: !Sub ${AWS::StackName}-public-api
```
sandymay-dd, line 1145: this API doesn't exist anymore - the name has changed

backend-api/template.yaml
```diff
@@ -1112,6 +1112,126 @@ Resources:
...
```
sandymay-dd, line 1205: API name has changed in main!

backend-api/template.yaml
```diff
@@ -1112,6 +1112,126 @@ Resources:
         - !Ref PermissionsBoundary
         - !Ref AWS::NoValue
 
+  WellKnown5XXLowThresholdWarningAlarm:
```
sandymay-dd, line 1115: Please may we take out Warning from the alarm name

backend-api/template.yaml
```diff
@@ -1112,6 +1112,126 @@ Resources:
...
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      DatapointsToAlarm: 2
+      EvaluationPeriods: 5
+      TreatMissingData: notBreaching
+      Threshold: 2
+      Metrics:
+        - Id: errorThreshold
+          Label: errorThreshold
+          ReturnData: true
+          Expression: IF((invocations>4) && (errorSum5XX>2) && (errorPercentage5XX>1),errorPercentage5XX,0)
```
sandymay-dd, line 1134: As discussed I think these should be >4

sandymay-dd: I don't think we need this from the Expression `(errorSum5XX>2)` as this is calculated in line 1129 `(Threshold: 2)`

What do you think?

mthkbarber: My understanding is that the `threshold:2` should actually be `threshold:1` as this is the value that gets compared against the return value of expression which is `errorPercentage5XX` so I will rectify that.

However I think that `errorSum5XX` is also required because the criteria for this alarm is at least 2 5XX errors.

Which means in an example case of 10 invocations and 1 5XX (over 2 periods) error this would trigger the alarm as it would exceed the 1% threshold.

backend-api/template.yaml
mthkbarber: Changed to `${AWS::StackName}-sessions-api`

backend-api/template.yaml
mthkbarber: Changed to `${AWS::StackName}-sessions-api`

backend-api/template.yaml
```diff
@@ -1112,6 +1112,126 @@ Resources:
...
+                - Name: Resource
+                  Value: /.well-known/jwks.json
+            Period: 60
+            Stat: Sum
+        - Id: errorPercentage5XX
+          Label: "Number of 5XX errors returned as a percentage"
+          ReturnData: false
+          Expression: (errorSum5XX/invocations) * 100
+
+  WellKnown5XXHighThresholdCriticalAlarm:
```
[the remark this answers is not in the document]
mthkbarber: done

backend-api/template.yaml
```diff
@@ -1155,6 +1155,126 @@ Resources:
...
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      DatapointsToAlarm: 2
+      EvaluationPeriods: 5
+      TreatMissingData: notBreaching
+      Threshold: 1
+      Metrics:
+        - Id: errorThreshold
+          Label: errorThreshold
+          ReturnData: true
+          Expression: IF((invocations>=4) && (errorSum5XX>=2) && (errorPercentage5XX>=1),errorPercentage5XX,0)
```
sandymay-dd, line 1177:
```suggestion
          Expression: IF((invocations>=4) && (errorSum5XX>=2),errorPercentage5XX,0)
```

sandymay-dd: I don't think we need to evaluate the error % as that's being done in line 1172

backend-api/template.yaml
```diff
@@ -1155,6 +1155,126 @@ Resources:
...
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      DatapointsToAlarm: 2
+      EvaluationPeriods: 5
+      TreatMissingData: notBreaching
+      Threshold: 80
+      Metrics:
+        - Id: errorThreshold
+          Label: errorThreshold
+          ReturnData: true
+          Expression: IF(errorPercentage5XX>=80,errorPercentage5XX,0)
```
sandymay-dd, line 1237:
```suggestion
          Expression: IF(errorSum5XX>=10,errorPercentage5XX,0)
```

I think?

backend-api/tests/infra-tests/application.test.ts
```diff
@@ -331,6 +331,30 @@ describe("Backend application infrastructure", () => {
     });
   });
 
+  it("apiGateway WellKnown5XXHighThresholdAlarm should have a runbook if alarm is enabled", () => {
```
sandymay-dd, line 334: Please may we also have the low threshold alarm for the eventuality that someone marks it as critical by mistake?

sandymay-dd: Please may we add a new describe block for Alarms at the top level of this test file?

backend-api/template.yaml
```diff
@@ -1155,6 +1155,126 @@ Resources:
         - !Ref PermissionsBoundary
         - !Ref AWS::NoValue
 
+  WellKnown5XXLowThresholdAlarm:    
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      ActionsEnabled: true
+      AlarmDescription: "/.well-known/jwks.json endpoint has returned at least 2 5XX status codes twice in the past 5 minutes"
```
sandymay-dd, line 1162: It's not entirely clear from the ticket, but I think we should align this to the High Threshold alarm:

1% or more of requests are failing on the /.well-known/jwks.json endpoint

backend-api/template.yaml
mthkbarber: I think actually this should be `invocations>=10` as the ticket doesn't actually specify 10 errors? Unless we want to make an amendment?

PR #185, opened 2024-09-29
DCMAW-8389:  Part 1 API tests for client credentials grant flow

ahickmann-gds:
​DCMAW-8389

### What changed
- Update Open API specification
- Add API tests for `/token` and `/credential` endpoints via the proxy API
- Add a new secret **clientRegistryApiTest** to Secrets Manager to store registered clients for test purposes only
- In **Credential** **TokenService** make all KMS signature verification errors be client errors so that the lambda returns 400 when the signature cannot be verified instead of 500. **There is a [ticket](https://govukverify.atlassian.net/browse/DCMAW-10323) in the backlog for checking the KMS exception type and return server or client error accordingly** 
- In the **Token** handler, return 401 instead of 400 when the authorization header is missing or is malformed  

Screenshot 2024-10-03 at 14 20 30

### Why did it change
- To test the client credentials grant flow
- Correct error responses and match TD

## Checklists


backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -28,11 +36,41 @@ const interceptor = aws4Interceptor({
...
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_grant",
+        error_description: "Invalid grant type or grant type not specified",
+      });
+      expect(response.status).toBe(400);
+    });
+  });
+
+  describe("Given the grant_type value is invalid", () => {
```
ahickmann-gds, line 64: I'm not sure if these first two tests are both required as they return exactly the same response body.

sandymay-dd: I'm happy for these tests to serve as documentation and regression

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -59,4 +172,211 @@ describe("POST /credential", () => {
...
+        "https://vocab.account.gov.uk/v1/credentialStatus": "pending",
+        sub: "testSub",
+      });
+      expect(response.status).toBe(200);
+    });
+  });
+
+  describe("Given the request is valid but there is no active session", () => {
+    it("Returns 201 Created", async () => {
+      await deleteSessionIfExists(credentialRequestBody.sub);
```
sandymay-dd, line 263: I don't think we need this step to test the scenario - but let's chat

ahickmann-gds: @jamesamooney

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -59,4 +152,197 @@ describe("POST /credential", () => {
...
+      expect(response.data).toStrictEqual({
+        "https://vocab.account.gov.uk/v1/credentialStatus": "pending",
+        sub: randomSub,
+      });
+      expect(response.status).toStrictEqual(201);
+    });
+  });
 });
+
+interface ClientDetails {
```
ahickmann-gds, line 285: I can move these helpers somewhere else but I'm just a bit concerned about creating too many helpers/utils files...

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: Ok, keeping both then... thanks!

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: Removed as discussed twice.

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
sandymay-dd: if the helpers are for the file in question - let's keep them co-located. I despise a 'utils' folder - we can have a long chat on this one over a ☕ 😄

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,14 +58,110 @@ describe("POST /token", () => {
...
+        {
+          headers: {
+            "x-custom-auth": "Basic " + toBase64(clientIdAndSecret),
+          },
+        },
+      );
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_grant",
+        error_description: "Invalid grant type or grant type not specified",
```
ahickmann-gds, line 76: API spec says this `error_description`  should be `Invalid grant` - how do we want to proceed?

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,14 +58,110 @@ describe("POST /token", () => {
...
+        {
+          headers: {
+            "x-custom-auth": "Basic " + toBase64(invalidClientIdAndSecret),
+          },
+        },
+      );
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_client",
+        error_description: "Supplied client credentials not recognised",
```
ahickmann-gds, line 112: API spec says this `error_description` should be `Invalid client` - how do we want to proceed?

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -28,11 +33,21 @@ const interceptor = aws4Interceptor({
 axiosInstance.interceptors.request.use(interceptor);
 
 describe("POST /token", () => {
-  describe("Given there is no grant type", () => {
+  let clientIdAndSecret: string;
+
+  beforeAll(async () => {
+    const clientDetails = await getFirstClientDetails();
```
sandymay-dd, line 39: Just aligning on terminology with the lambda and services we've built:

The array in Secrets Manager = the Client Registry
An individual item in the array = a registeredClient

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +57,325 @@ describe("POST /token", () => {
...
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_client",
+        error_description: "Supplied client credentials not recognised",
+      });
+      expect(response.status).toBe(400);
+    });
+  });
+
+  describe("Given the request is valid", () => {
```
sandymay-dd, line 117: ... and the client is registered just to align with the test above?

backend-api/src/functions/tests/pact/pactFiles/IpvCoreBack-DcmawAsyncCriProvider.json
```diff
@@ -51,7 +51,7 @@
         "path": "/async/credential"
       },
       "response": {
-        "status": 400
+        "status": 401
```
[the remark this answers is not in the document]
ahickmann-gds: Reverted...

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: It's just that we are not getting this information from the client registry - we had to create a new secret and for this reason I went for client details (i.e. to make it clearer that this data isn't coming from the client registry)

Maybe I can call it testClientRegistry instead...

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: Done

jmooney-dd:
Should we ignore `apiInstance.ts` from coverage reports on Sonar?

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +59,325 @@ describe("POST /token", () => {
...
+  describe("Given there is no Authorization header in the request", () => {
+    it("Returns a 400 Bad Request response", async () => {
+      const response = await axiosInstance.post(
+        `${apiBaseUrl}/async/token`,
+        "grant_type=client_credentials",
+      );
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_client",
+        error_description: "Supplied client credentials not recognised",
```
sandymay-dd, line 92: I think this one should be a 401?

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +59,325 @@ describe("POST /token", () => {
...
+            "X-Custom-Auth": "Bearer " + accessTokenWithInvalidSignature,
+          },
+        },
+      );
+
+      expect(response.data).toStrictEqual({
+        error: "server_error",
+        error_description: "Server Error",
+      });
+      expect(response.status).toBe(500);
```
sandymay-dd, line 237: Nice! It's often v hard to check 500 responses

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: Possibly, that's what the TD says, but the lambda never returned a 401 even though that was in the spec. Should we make the lambda return 401 when the header is missing and 400 when the header is not valid? 😕

backend-api/openApiSpecs/async-private-spec.yaml
```diff
@@ -100,34 +120,44 @@ paths:
...
-                    The authorisation code is invalid or expired. This is also the error which would return if the redirect URL given in the authorisation request does not match the URL provided in this access token request.
+                    The grant type is missing or the value provided is not supported by the server.
                   value:
                     {
                       "error": "invalid_grant",
                       "error_description": "Invalid grant"
                     }
               schema:
-                $ref: "#/components/schemas/Error"
-        '401':
```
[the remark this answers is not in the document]
ahickmann-gds: What does it mean if we support 401 but our code never returns 401?

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -1,8 +1,15 @@
...
-
 import "dotenv/config";
+import {
+  GetSecretValueCommand,
+  SecretsManagerClient,
+} from "@aws-sdk/client-secrets-manager";
+import { randomUUID } from "crypto";
+import { UUID } from "node:crypto";
+
+process.env.TEST_ENVIRONMENT = "dev";
```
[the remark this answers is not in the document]
ahickmann-gds: It does - I keep forgetting and adding it back. It was meant to be just for me for testing locally...

backend-api/openApiSpecs/async-private-spec.yaml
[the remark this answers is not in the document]
ahickmann-gds: Returning 401 now.

backend-api/src/functions/asyncToken/asyncTokenHandler.ts
```diff
@@ -38,15 +36,15 @@ export async function lambdaHandlerConstructor(
     logger.log("INVALID_REQUEST", {
       errorMessage: eventBodyResult.value.errorMessage,
     });
-    return badRequestResponseInvalidGrant;
+    return badRequestResponse("Invalid grant type or grant type not specified");
```
sandymay-dd, line 39: I almost think this would look nice as a builder. So you don't have to scroll down to work out what badRequestResponse is. 

Not for now but would be curious for your thoughts as a talking point

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +76,322 @@ describe("POST /token", () => {
...
+            "X-Custom-Auth": "Bearer invalid.access.token",
+          },
+        },
+      );
+
+      expect(response.data).toStrictEqual({
+        error: "invalid_token",
+        error_description: "JWT payload not valid JSON",
+      });
+      expect(response.status).toBe(400);
```
sandymay-dd, line 189: I think this should be a 401?

sandymay-dd: hmmm ignore me - I think it's in a 'valid' format

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +76,322 @@ describe("POST /token", () => {
...
+      expect(responseOne.status).toBe(200);
+      expect(responseTwo.data).toStrictEqual({
+        "https://vocab.account.gov.uk/v1/credentialStatus": "pending",
+        sub: "urn:fdc:gov.uk:2022:56P4CMsGh_02YOlWpd8PAOI-2sVlB2nsNU7mcLZYhYw=",
+      });
+      expect(responseTwo.status).toBe(200);
+    });
+  });
+
+  describe("Given the request is valid and a new session is created for a new sub", () => {
```
sandymay-dd, line 306:
```suggestion
  describe("Given the request is valid and there is no active session for the given sub", () => {
```

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -78,13 +78,7 @@ export async function lambdaHandlerConstructor(
   );
   if (verifyTokenSignatureResult.isError) {
     const errorMessage = verifyTokenSignatureResult.value.errorMessage;
-    if (verifyTokenSignatureResult.value.errorCategory === "SERVER_ERROR") {
```
sandymay-dd, line 81: I'm sorry - please may you help me work out why this can't return a server error? Slow brain

backend-api/src/functions/asyncCredential/tokenService/tokenService.ts
```diff
@@ -41,19 +41,11 @@ export class TokenService implements IDecodeToken, IVerifyTokenSignature {
...
       return errorResult({
-        errorMessage: "Unexpected error when verifying jwt signature with KMS",
-        errorCategory: "SERVER_ERROR",
-      });
-    }
-
-    if (result.SignatureValid !== true) {
-      return errorResult({
-        errorMessage: "Signature is invalid",
+        errorMessage: "Signature could not be verified",
```
sandymay-dd, line 48: Yep I think this function can still return a server error 👀

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
ahickmann-gds: Renamed :)

sandymay-dd (approved):
Nice work - as discussed some follow up work on the credentials lambda's token service to see whether we can do this without using KMS - out of scope for this change.

ahickmann-gds:
> Nice work - as discussed some follow up work on the credentials lambda's token service to see whether we can do this without using KMS - out of scope for this change.

Agreed! I will raise a ticket to put in the backlog.

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -42,21 +59,325 @@ describe("POST /token", () => {
...
+      expect(response.data).toStrictEqual({
+        "https://vocab.account.gov.uk/v1/credentialStatus": "pending",
+        sub: "urn:fdc:gov.uk:2022:56P4CMsGh_02YOlWpd8PAOI-2sVlB2nsNU7mcLZYhYw=",
+      });
+      expect(response.status).toBe(200);
+    });
+  });
+
+  describe("Given the same access token is used more than once to fetch an active session", () => {
+    it("Returns 200 OK", async () => {
```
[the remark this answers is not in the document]
ahickmann-gds: Very good idea - done!

PR #183, opened 2024-09-26
DCMAW-8229: Fetch public key and get content encryption key

jhumbert-dd:
​DCMAW-8229

### What changed
- Add environment variables
  - `ENCRYPTION_KEY_ARN`
  - `STS_JWKS_ENDPOINT`
- Add `ConfigService`
- Add `TokenService` which currently:
  - fetches the STS public key (to verify token signature in a following PR)
  - gets and decode Content Encryption Key (CEK) from Authorization header
 - Add `KmsAdapter` for decrypting the CEK
   - Add permission for `kms:decrypt` to `AsyncActiveSessionLambdaRole`

### Why did it change

This is part of the work to link the IPV session started in the browser with the DCMAW session and resume the journey within the mobile app.

### Small follow-up PRs will include
- Retry policy when hitting STS's Jwks endpoint
- Caching public key
- Updating auth header log to be common

### Evidence

- Deployed a custom STS mock that points towards our custom BE API. 
- Hit custom STS mock with a valid request to get back a service token:
- Use this token to hit our /async/activeSession endpoint and we get a 200 response back

Request to STS token
Screenshot 2024-09-27 at 15 25 23

Request to `async/activeSession`
Screenshot 2024-09-27 at 15 22 45

## Checklists

- [x] There is a ticket raised for this PR that is present in the branch name
- [x] No PII data logged. [See guidance here](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3502407722/PII+Logging+Considerations)
- [ ] Demo to a BA, TA, and the team.
- [ ] Update [README](./blob/main/README.md) with any new instructions or tasks

PR #186, opened 2024-09-30
DCMAW-8229: Update logic in shared util in request service

jmooney-dd:
​DCMAW-8229

### What changed
- Updated request service util file and function name to be more specific to what it does
- Updates variable name inside util func, `jwe` now is `token`
- Updates imports
- Updates tokenService to not split header as this is already done in util func
- Updates tokenService tests as func is no longer expecting token starting with the word bearer

### Why did it change
In #183 we changed how a shared util works by returning the token already being split (i.e. not returning the token with the word Bearer at the start). The token service used by the Credential lambda was not expecting this and therefore started returning 500 responses.

This util function is used in the `asyncCredential` and `asyncActiveSession` lambdas.

### Evidence

Previously when calling credential endpoint we'd receive a 500. Ana found the following log:

Screenshot 2024-09-30 at 15 34 52

Deployed new stack and now we get a success response again:

Screenshot 2024-09-30 at 15 35 27

Also confirmed that activeSession endpoint is still working:

Screenshot 2024-09-30 at 15 43 22

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncActiveSession/requestService/requestService.ts
+++ b/backend-api/src/functions/asyncActiveSession/requestService/requestService.ts
@@ -1,9 +1,9 @@
-import { getBearerTokenFromHeader } from "../../services/utils/requestService";
+import { getBearerToken } from "../../services/utils/getBearerToken";
 import { Result } from "../../utils/result";
 
 export class RequestService implements IRequestService {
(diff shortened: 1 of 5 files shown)


ahickmann-gds (commented):
Thank you! :)

ahickmann-gds (approved):

PR #187, opened 2024-09-30
DCMAW-8229: Adds retry policy to activeSession token service STS jwks call

jmooney-dd:
​DCMAW-XXXXX

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
+++ b/backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
@@ -127,26 +127,26 @@ describe("Token Service", () => {
           describe("Given there is an error retrieving the public key on the first attempt", () => {
             it("Makes second attempt to get STS key", async () => {
               mockFetch = jest
-              .spyOn(global, "fetch")
-              .mockImplementationOnce(() =>
-                Promise.resolve({
-                  status: 200,
-                  ok: true,
-                  json: () =>
-                    Promise.resolve({
-                      keys: [
-                        {
-                          kty: "mockKty",
-                          x: "mockX",
-                          y: "mockY",
-                          crv: "mockCrv",
-                          d: "mockD",
-                          kid: "mockKid",
-                        },
-                      ],
-                    }),
-                } as Response),
-              )
+                .spyOn(global, "fetch")
+                .mockImplementationOnce(() =>
+                  Promise.resolve({
+                    status: 200,
+                    ok: true,
+                    json: () =>
+                      Promise.resolve({
+                        keys: [
+                          {
+                            kty: "mockKty",
+                            x: "mockX",
+                            y: "mockY",
+                            crv: "mockCrv",
+                            d: "mockD",
+                            kid: "mockKid",
+                          },
+                        ],
+                      }),
+                  } as Response),
+                );
 
               await tokenService.getSubFromToken(
                 "https://mockJwksEndpoint.com",
@@ -159,34 +159,36 @@ describe("Token Service", () => {
                   method: "GET",
                 },
               );
-              expect(mockFetch).toHaveBeenCalledTimes(2)
-            })
-          })
+              expect(mockFetch).toHaveBeenCalledTimes(1);
+            });
+          });
 
           describe("Given there is an error retrieving the public key on the second attempt", () => {
             it("Makes third attempt to get STS key", async () => {
               mockFetch = jest
-              .spyOn(global, "fetch")
-              .mockImplementationOnce(() => Promise.reject(new Error("mockError")))
-              .mockImplementationOnce(() =>
-                Promise.resolve({
-                  status: 200,
-                  ok: true,
-                  json: () =>
-                    Promise.resolve({
-                      keys: [
-                        {
-                          kty: "mockKty",
-                          x: "mockX",
-                          y: "mockY",
-                          crv: "mockCrv",
-                          d: "mockD",
-                          kid: "mockKid",
-                        },
-                      ],
-                    }),
-                } as Response),
-              )
+                .spyOn(global, "fetch")
+                .mockImplementationOnce(() =>
+                  Promise.reject(new Error("mockError")),
+                )
+                .mockImplementationOnce(() =>
+                  Promise.resolve({
+                    status: 200,
+                    ok: true,
+                    json: () =>
+                      Promise.resolve({
+                        keys: [
+                          {
+                            kty: "mockKty",
+                            x: "mockX",
+                            y: "mockY",
+                            crv: "mockCrv",
+                            d: "mockD",
+                            kid: "mockKid",
+                          },
+                        ],
+                      }),
+                  } as Response),
+                );
 
               await tokenService.getSubFromToken(
                 "https://mockJwksEndpoint.com",
@@ -199,17 +201,23 @@ describe("Token Service", () => {
                   method: "GET",
                 },
               );
-              expect(mockFetch).toHaveBeenCalledTimes(3)
-            })
-          })
+              expect(mockFetch).toHaveBeenCalledTimes(2);
+            });
+          });
 
           describe("Given there is an error retrieving the public key on the third attempt", () => {
             it("Returns error result", async () => {
               mockFetch = jest
-              .spyOn(global, "fetch")
-              .mockImplementationOnce(() => Promise.reject(new Error("mockError")))
-              .mockImplementationOnce(() => Promise.reject(new Error("mockError")))
-              .mockImplementationOnce(() => Promise.reject(new Error("mockError")))
+                .spyOn(global, "fetch")
+                .mockImplementationOnce(() =>
+                  Promise.reject(new Error("mockError")),
+                )
+                .mockImplementationOnce(() =>
+                  Promise.reject(new Error("mockError")),
+                )
+                .mockImplementationOnce(() =>
+                  Promise.reject(new Error("mockError")),
+                );
 
               const result = await tokenService.getSubFromToken(
                 "https://mockJwksEndpoint.com",
@@ -222,16 +230,16 @@ describe("Token Service", () => {
                   method: "GET",
                 },
               );
-              expect(mockFetch).toHaveBeenCalledTimes(3)
+              expect(mockFetch).toHaveBeenCalledTimes(3);
 
               expect(result.isError).toBe(true);
               expect(result.value).toStrictEqual({
                 errorMessage: "Unexpected error retrieving STS public key",
                 errorCategory: "SERVER_ERROR",
               });
-            })
-          })
-        })
+            });
+          });
+        });
       });
     });
 
--- a/backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
+++ b/backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
@@ -122,6 +122,116 @@ describe("Token Service", () => {
             });
           });
         });
+
+        describe("Retry policy", () => {
+          describe("Given there is an error retrieving the public key on the first attempt", () => {
+            it("Makes second attempt to get STS key", async () => {
+              mockFetch = jest
+              .spyOn(global, "fetch")
+              .mockImplementationOnce(() =>
+                Promise.resolve({
+                  status: 200,
+                  ok: true,
+                  json: () =>
+                    Promise.resolve({
+                      keys: [
+                        {
+                          kty: "mockKty",
+                          x: "mockX",
+                          y: "mockY",
+                          crv: "mockCrv",
+                          d: "mockD",
+                          kid: "mockKid",
+                        },
+                      ],
+                    }),
+                } as Response),
+              )
+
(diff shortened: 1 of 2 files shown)


backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
```diff
@@ -122,6 +122,130 @@ describe("Token Service", () => {
             });
           });
         });
+
+        describe("Retry policy", () => {
```
jmooney-dd, line 126: We should use jest fake timers so we don't actually have to wait the 'real' delay for each retry

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
```diff
@@ -52,6 +51,38 @@ export class TokenService implements ITokenService {
...
+      }
+
+      if (retries < maxRetries) {
+        await new Promise((resolve) => setTimeout(resolve, delayInMs));
+      }
+    }
+
+    // After all retries, check if fetchPublicKeyResult was assigned and handle the case where it was not assigned (should not happen)
+    if (fetchPublicKeyResult == null) {
+      return errorResult({
```
jmooney-dd, line 76: Not tested but not sure it's worth as it shouldn't ever happen, this is to satisify a satisfy. Does anyone:
- Know a way to avoid this check?
- If not, think we should still unit test it? Would we have to mock the `fetchPublicKey` func (which is currently readonly)

backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
jmooney-dd: Or delay could be an arg taken by the func that we could modify if jest timers don't play kindly with calls like this. Will investigate tomorrow

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
```diff
@@ -52,6 +51,38 @@ export class TokenService implements ITokenService {
     return successResult("");
   };
 
+  private async fetchPublicKeyWithRetries(
+    stsJwksEndpoint: string,
+  ): Promise<Result<JSON>> {
+    const maxRetries = 2;
+    const delayInMs = 1000;
```
sandymay-dd, line 58: Not looking at this in depth but I'd suggest if we were using a linear retry strategy by waiting the same amount each time, I'd recommend a 100ms delay as a starting point

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
sandymay-dd: Thinking about the above ^ I'd almost recommend casting the .json() response to `unknown`. This means the compiler will explicitly force us to check the structure of the response. `any` is a way of telling the compiler: 'Trust me - I know what I'm doing and I can perform any method on this mystery variable'

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
```diff
@@ -52,6 +51,38 @@ export class TokenService implements ITokenService {
...
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+
+    return fetchPublicKeyResult;
+  }
+
   private readonly fetchPublicKey = async (
     stsJwksEndpoint: string,
   ): Promise<Result<JSON>> => {
```
ahickmann-gds, line 88: I think we can't say the type is JSON because we don't know that it actually is as `.json` returns `any` 🤔

jmooney-dd: Feels counter intuitive but would it therefore be better to do response.text() and then parse it so we know it's JSON?

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: Feels counter intuitive but would it be better to do response.text() and then parse it so we know it's JSON? Or do we make a public key type that we validate it's the shape of?

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: Or do we make a public key type that we validate it's the shape of?

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: The TD states _The retries should occur in intervals of 1 second_ which is why we did the above. Are you saying you'd prefer the first retry to occur after 100ms and the second to occur after 1000ms?

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: After discussion will validate the keys

backend-api/src/functions/adapters/fetchAdapter.ts
```diff
@@ -0,0 +1,7 @@
+export const fetchAdapter: IFetchAdapter = (url, options) => {
```
jmooney-dd, line 1: Not how this implementation will work. Aiming to somewhat replicate functionality here but without exponential back off: https://github.com/govuk-one-login/sts-back/blob/main/backend-api/src/common/http/sendHttpRequest.ts

Can I check before I do this that you're happy with this @sandymay-dd re the retry strategy. If we will eventually want exponential back off here then might be easy to lift the entire func instead of refactoring to be simpler. If we do just want a linear retry strat then I will do the refactor 

cc: @ahickmann-gds

jmooney-dd: FYI currently the retry stuff is in the token service

jmooney-dd: Am doing a simplified version FYI.

backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
```diff
@@ -122,6 +125,167 @@ describe("Token Service", () => {
...
+                  Promise.resolve({
+                    keys: ["mockNotAnObject"],
+                  }),
+              } as Response),
+            );
+
+            const result = await tokenService.getSubFromToken(
+              "https://mockJwksEndpoint.com",
+              "mockJwe",
+              100,
```
jmooney-dd, line 145: Should I make this even smaller for the tests? Could do 10 or even 1ms delay to speed up unit tests?

backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -0,0 +1,96 @@
...
+        body,
+      });
+
+      if (!response.ok && attempt < maxAttempts) {
+        return retry(request, delayInMillis);
+      }
+
+      if (!response.ok) {
+        return errorResult({
+          errorMessage: "Error retrieving STS public keys",
```
ahickmann-gds, line 33: I'd make these two error messages on lines 33 and 43 more generic (i.e. not mention STS public keys)

backend-api/template.yaml
```diff
@@ -789,6 +789,7 @@ Resources:
           ENCRYPTION_KEY_ARN: !GetAtt KMSEncryptionKey.Arn
           STS_JWKS_ENDPOINT:
             !FindInMap [EnvironmentVariables, !Ref Environment, STSJWKSENDPOINT]
+          STS_JWKS_ENDPOINT_RETRY_DELAY_IN_MS: 1000
```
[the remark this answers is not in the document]
ahickmann-gds: I'm not saying it can't be an env var or that this is wrong, I'm just wondering if we need one...

ahickmann-gds, line 792: I made comments about `STS_JWKS_ENDPOINT_RETRY_DELAY_IN_MS` being an env var and that was silly me, so I've deleted them. Do we want maxAttempts to be an env var as well?

backend-api/src/functions/services/http/sendHttpRequest.ts
jmooney-dd: Updated

backend-api/template.yaml
jmooney-dd: Done

backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -0,0 +1,96 @@
...
+        return retry(request, delayInMillis);
+      }
+
+      return errorResult({
+        errorMessage: "Unexpected error retrieving STS public keys",
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+
+    return successResult(response);
```
ahickmann-gds, line 48: I'd return the response as string so that you don't need to make "fetch specific" calls outside this function
```suggestion
    return successResult(await response.text());
```

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
```diff
@@ -43,75 +56,74 @@ export class TokenService implements ITokenService {
...
+    return successResult(cek);
+  }
+
+  private async getJwksFromResponse(
+    response: Response,
+  ): Promise<Result<IJwks>> {
+    let jwks;
     try {
-      publicKey = await response.json();
+      jwks = await response.json();
```
ahickmann-gds, line 103: If you return the response as string then here you could use `JSON.parse()`

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
```diff
@@ -43,75 +59,74 @@ export class TokenService implements ITokenService {
...
-      response = await fetch(stsJwksEndpoint, {
-        method: "GET",
-      });
-    } catch {
-      return errorResult({
-        errorMessage: "Unexpected error retrieving STS public key",
-        errorCategory: "SERVER_ERROR",
-      });
+  private readonly isJwks = (data: unknown): data is IJwks => {
+    return (
```
ahickmann-gds, line 69: Could this be simplified? 
```suggestion
  return (
      typeof data === "object" &&
      data !== null &&
      "keys" in data &&
      Array.isArray(data.keys) &&
      data.keys.every((key) => typeof key === "object" && key !== null)
    );
```

ahickmann-gds: I like `isJwks` 😄

jmooney-dd: Thanks! updated

backend-api/src/functions/services/http/sendHttpRequest.ts
jmooney-dd: Updated to return a success http response type now

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: Updated

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -38,6 +38,10 @@ export async function lambdaHandlerConstructor(
   const getSubFromTokenResult = await tokenService.getSubFromToken(
     config.STS_JWKS_ENDPOINT,
     serviceToken,
+    {
+      maxAttempts: 3,
+      delayInMillis: 1000,
```
sandymay-dd, line 43: Please may we keep this at 100ms

backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -0,0 +1,106 @@
...
+async function wait(delayMillis: number): Promise<void> {
+  await new Promise((resolve) => setTimeout(resolve, delayMillis));
+}
+
+export type RetryConfig = {
+  maxAttempts?: number;
+  delayInMillis?: number;
+};
+
+export type HttpMethod =
```
sandymay-dd, line 75: Please may we only support the methods we need for now?

backend-api/src/functions/asyncActiveSession/tokenService/tokenService.ts
jmooney-dd: The above code is outdated but the question is still the same

sandymay-dd: I think typo - this is an excessive amount of time to wait. Please may we update to 100ms and I'll update the TD now

jmooney-dd: Gotcha! Will update now

backend-api/src/functions/services/http/sendHttpRequest.ts
jmooney-dd: Done

ahickmann-gds (approved):

PR #192, opened 2024-10-01
DCMAW-8229: KMS adapter

ahickmann-gds:
​DCMAW-8229

### What changed
- Make the `decrypt` function accept the key ARN
- Return error result if `Plaintext` in the output of the decrypt command is `null`
- Import `KmsAdapter` in the `TokenService` instead of making it a property of that class

Screenshot 2024-10-01 at 14 12 05

## Checklists


backend-api/src/functions/adapters/kmsAdapter.ts
```diff
@@ -1,50 +1,70 @@
...
-          CiphertextBlob: encryptedCek,
+          KeyId: keyArn,
+          CiphertextBlob: ciphertext,
           EncryptionAlgorithm: "RSAES_OAEP_SHA_256",
         }),
       );
-    } catch {
+    } catch (error) {
+      if (
+        error instanceof InvalidCiphertextException ||
```
ahickmann-gds, line 38: I don't think we can use the exception property `$fault` to decide whether we return a client or a server error - we can talk about this over a call - so we need to check the exception type. I think we need to have a discussion about how we want to handle the different exceptions KMS can throw and if we want to return CLIENT_ERROR or something else from the function.

backend-api/src/functions/adapters/kmsAdapter.ts
```diff
@@ -1,50 +1,70 @@
 import {
   DecryptCommand,
   DecryptCommandOutput,
+  IncorrectKeyException,
+  InvalidCiphertextException,
   KMSClient,
 } from "@aws-sdk/client-kms";
-import { NodeHttpHandler } from "@smithy/node-http-handler";
 import { errorResult, Result, successResult } from "../utils/result";
+import { NodeHttpHandler } from "@aws-sdk/node-http-handler";
```
jmooney-dd, line 9: As we're using the aws sdk v3 I think the smithy package is the preferred option (what we had before). Both work of course!

backend-api/src/functions/adapters/kmsAdapter.ts
ahickmann-gds: Yes. When we call KMS, we are the client and they are the server, so the fault property (when set to `client`) is saying that we as their client made a mistake, but that does not mean that our client made a mistake. 🫤

ahickmann-gds: Re your first question, the KMS `decrypt` method throws various exceptions. We don't need to check every single one of them in the code but we should probably go through the list of possible exceptions and decide which ones should be a client error and which should be server error. Or maybe we don't want to check at all and just return a server error!

jmooney-dd: Yeah I'm thinking a blanket server error actually might be the way to go here. I'm trying to think of the value of checking but can't think of much

backend-api/src/functions/adapters/kmsAdapter.ts
ahickmann-gds: I didn't mean to change this, it was a mistake, sorry.

jmooney-dd (approved):

backend-api/src/functions/adapters/kmsAdapter.ts
ahickmann-gds: I've stopped checking the KMS exception type and instead chose to catch all errors and return a SERVER_ERROR like before just so that we can merge this PR. However, I do think we need to have a conversation about error handling because right now I feel that we are swallowing these errors and we can't tell what's going wrong. :confusing

jmooney-dd (approved):

PR #193, opened 2024-10-01
DCAMW-8229: Add Lambdas to VPC

jhumbert-dd:
​DCMAW-8229

### What changed
- Add `VpcConfig` to Lambda functions and `VpcPolicy` to Lambda IAM Roles for:
  - AsyncToken
  - AsyncCredential
  - AsyncActiveSession
  - Jwks
- Add infra tests

Async backend and STS mock stacks were deployed to dev and tested:

Token & Credential
Screenshot 2024-10-02 at 19 54 20

STS Mock Token & Active Session
Screenshot 2024-10-02 at 19 55 24

### Why did it change
- It is a security best practice to restrict access to functions

#### All tests passing using `run-tests-locally.sh` script against custom deployed   stack:

Screenshot 2024-10-01 at 14 44 15
Screenshot 2024-10-01 at 14 44 32
Screenshot 2024-10-01 at 14 44 51

## Checklists

### DIFF
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -372,13 +372,13 @@ Resources:
         Variables:
           CLIENT_REGISTRY_SECRET_NAME: !Sub ${Environment}/clientRegistry
           SESSION_DURATION_IN_SECONDS: 3600 #Used to set time to live when creating sessions. Set to 1 hour.
-      # VpcConfig:
-      #   SubnetIds:
-      #     - !ImportValue devplatform-vpc-PrivateSubnetIdA
-      #     - !ImportValue devplatform-vpc-PrivateSubnetIdB
-      #     - !ImportValue devplatform-vpc-PrivateSubnetIdC
-      #   SecurityGroupIds:
-      #     - !ImportValue devplatform-vpc-AWSServicesEndpointSecurityGroupId
+      VpcConfig:
+        SubnetIds:
+          - !ImportValue devplatform-vpc-PrivateSubnetIdA
+          - !ImportValue devplatform-vpc-PrivateSubnetIdB
+          - !ImportValue devplatform-vpc-PrivateSubnetIdC
+        SecurityGroupIds:
+          - !ImportValue devplatform-vpc-AWSServicesEndpointSecurityGroupId
 
   AsyncCredentialLogGroup:
     Type: AWS::Logs::LogGroup
@@ -398,6 +398,7 @@ Resources:
   AsyncCredentialLambdaRole:
     Type: AWS::IAM::Role
     Properties:
+      RoleName: !Sub ${AWS::StackName}-credential-lambda
       AssumeRolePolicyDocument:
         Version: "2012-10-17"
         Statement:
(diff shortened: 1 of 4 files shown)


backend-api/template.yaml
```diff
@@ -327,7 +335,16 @@ Resources:
                   - kms:Decrypt
                   - kms:GenerateDataKey
                 Resource: !GetAtt TxMAKMSEncryptionKey.Arn
-      RoleName: !Sub ${AWS::StackName}-token-lambda
+        - PolicyName: ENIPolicy
```
jhumbert-dd, line 338: Does this name makes sense for the modified lambda roles in this PR? @sandy-may

jhumbert-dd, line 338: This could have been done another way using `ManagedPolicyArns`, which would reduce the size of the template but maybe adds more permissions than needed. Any opinions on which option would work best here?

ahickmann-gds: Maybe VpcConfigurationPolicy / VpcConfigPolicy / VpcAccessPolicy?

ahickmann-gds (commented):
Can we also add the `JsonWebKeysFunction` to the VPC (ProtectedSubnet)? I forgot to mention this before, sorry.

backend-api/template.yaml
```diff
@@ -259,6 +259,13 @@ Resources:
       Environment:
         Variables:
           CLIENT_REGISTRY_SECRET_NAME: !Sub ${Environment}/clientRegistry
+      VpcConfig:
+        SubnetIds:
+          - !ImportValue devplatform-vpc-PrivateSubnetIdA
+          - !ImportValue devplatform-vpc-PrivateSubnetIdB
+          - !ImportValue devplatform-vpc-PrivateSubnetIdC
+        SecurityGroupIds:
+          - !ImportValue devplatform-vpc-AWSServicesEndpointSecurityGroupId
```
ahickmann-gds, line 268: If we need to create our own security group instead of referencing this one (i.e. `AWSServicesEndpointSecurityGroupId`), can we perhaps rename the `ProxyLambdaSecurityGroup` to something like `LambdaEgressSecurityGroup` and reference this security group throughout this template instead of `AWSServicesEndpointSecurityGroupId`? This existing security group seems to be suitable. cc @sandymay-dd

backend-api/template.yaml
```diff
@@ -259,6 +259,13 @@ Resources:
       Environment:
         Variables:
           CLIENT_REGISTRY_SECRET_NAME: !Sub ${Environment}/clientRegistry
+      VpcConfig:
```
sandymay-dd, line 262: Please may we write an infra test that checks all lambdas have a VpcConfig with 3 SubnetIds and a Security group attached?

backend-api/template.yaml
```diff
@@ -127,9 +127,9 @@ Mappings:
 
   EnvironmentVariables:
     dev:
-      STSJWKSENDPOINT: 'https://mob-sts-mock.review-b-async.dev.account.gov.uk/.well-known/jwks.json'
+      STSJWKSENDPOINT: 'https://sts-mock-ah.review-b-async.dev.account.gov.uk/.well-known/jwks.json'
```
sandymay-dd, line 130: Assume this isn't out for full review yet :)

ahickmann-gds: Assumed correctly 😄 

Fixed!

backend-api/template.yaml
ahickmann-gds: Done, but please check if ok :)

PR #197, opened 2024-10-03
DCMAW-8229: DynamoDB refactor

ahickmann-gds:
​DCMAW-8229

## Checklists


backend-api/src/functions/adapters/kmsAdapter.ts
```diff
@@ -7,16 +7,14 @@ import { errorResult, Result, successResult } from "../utils/result";
...
-      requestHandler: new NodeHttpHandler({
-        connectionTimeout: 29000,
-        requestTimeout: 29000,
-      }),
-    },
-  ]);
+  private readonly kmsClient = new KMSClient({
+    region: "eu-west-2",
+    maxAttempts: 2,
+    requestHandler: new NodeHttpHandler({
```
sandymay-dd, line 13: I think we need to have a think about this! The lambda timeout is under this threshold, so we'll never hit this 29000 value

ahickmann-gds: I will revert this change and fix it in another PR

backend-api/template.yaml
```diff
@@ -428,7 +428,6 @@ Resources:
                     dynamodb:Select: SPECIFIC_ATTRIBUTES
               - Effect: Allow
                 Action:
-                  - dynamodb:GetItem
```
ahickmann-gds, line 431: Getting the session before creating it is no longer required. We can use a condition expression to ensure that a new session is only created if one with the same session ID does not already exist.

backend-api/src/functions/services/sessionService/tests/sessionService.test.ts
```diff
@@ -0,0 +1,300 @@
...
+            issuer: "mockIssuer",
+            sessionDurationInSeconds: 12345,
+          });
+
+          expect(result.isError).toBe(false);
+          expect(result.value).toEqual(expect.any(String));
+        });
+      });
+
+      describe("Given the function input contains the key redirect_uri", () => {
```
ahickmann-gds, line 280: I am not entirely sure this test is even needed

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+      connectionTimeout: 29000,
+      requestTimeout: 29000,
+    }),
+  });
+
+  constructor(tableName: string) {
+    this.tableName = tableName;
+  }
+
+  async readSessionId(
```
sandymay-dd, line 35: please may we update this verb to be something like 'fetch' or 'get'

sandymay-dd: or perhaps 'getSessionIdFromSub'

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+    subjectIdentifier: string,
+  ): Promise<Result<string | null>> {
```
sandymay-dd, line 37: in which instance could this be null? I might be missing something

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+          errorMessage: "Unexpected error while creating a new session",
+          errorCategory: "SERVER_ERROR",
+        });
+      }
+    }
+
+    return successResult(sessionId);
+  }
+
+  private async query(
```
sandymay-dd, line 176: are we ok to move this up a little to where it's first used?

sandymay-dd, line 176: just on naming - perhaps sendQueryCommand?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: When there is no active session

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+        errorMessage: "Unexpected error when querying database",
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+
+    if (!output.Items || output.Items.length === 0) {
+      return successResult(null);
+    }
+
+    return successResult(output.Items[0]);
```
sandymay-dd, line 193: what happens if there are two items in the array - or is this scenario not expected? This feels like logic that should live in the function that returned. This function is now doing two things - making the query and choosing to return the first item

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+    };
+
+    if (redirect_uri) {
+      input.Item!.redirectUri = { S: redirect_uri };
+    }
+
+    try {
+      await this.dynamoDbClient.send(new PutItemCommand(input));
+    } catch (error) {
+      if (error instanceof ConditionalCheckFailedException) {
```
sandymay-dd, line 160: Yes! Well done

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+        sessionState: { S: "ASYNC_AUTH_SESSION_CREATED" },
+        state: { S: state },
+        subjectIdentifier: { S: sub },
+        timeToLive: { N: timeToLive.toString() },
+      },
+      ConditionExpression: "attribute_not_exists(sessionId)",
+    };
+
+    if (redirect_uri) {
+      input.Item!.redirectUri = { S: redirect_uri };
```
sandymay-dd, line 154: it feels very strange you have to have the `!` 🤔

backend-api/src/functions/adapters/dynamoDbAdapter.ts
sandymay-dd: Thank you. I misread this block:

`if (!sessionId) {
      return errorResult({
        errorMessage: "Session is malformed",
      });
    }`

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: You are right. The two query inputs we have right now do not allow two items to be returned because they both specify one item only, so taking into account how we are using this `query` method now, returning the first item is not a problem. However, you are absolutely right that it should not return output.Items[0] even if it can - it's not clear at all and makes it impossible to re-use the method with a different type of query input. I will fix that... 👀

backend-api/src/functions/asyncCredential/sessionService/dynamoDbClient.ts
sandymay-dd, line 1: please may we also delete this file 😅

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+    const currentTimeInSeconds = Math.floor(Date.now() / 1000);
+    const input: QueryCommandInput = {
+      TableName: this.tableName,
+      IndexName: "subjectIdentifier-timeToLive-index",
+      KeyConditionExpression:
+        "subjectIdentifier = :subjectIdentifier and :currentTimeInSeconds < timeToLive",
+      FilterExpression: "sessionState = :sessionState",
+      ExpressionAttributeValues: {
+        ":subjectIdentifier": { S: subjectIdentifier },
+        ":sessionState": { S: "ASYNC_AUTH_SESSION_CREATED" },
```
sandymay-dd, line 47: Thinking ahead a bit here - we're going to soon start getting the session record when looking up via the sessionId

I'm keen not to hardcode this string as we'll soon forget it's hardcoded.

Perhaps we should have some form of constants file in the sessionService folder?

const sessionStates = {
    ASYNC_AUTH_SESSION_CREATED: "ASYNC_AUTH_SESSION_CREATED"
} as const

or something similar

The same for when it's used in the next function

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+        ":currentTimeInSeconds": {
+          N: currentTimeInSeconds.toString(),
+        },
+      },
+      ProjectionExpression: "sessionId",
+      Limit: 1,
```
sandymay-dd, line 53: Does this retrieve the most recent if there are two sessions that fit this criteria (that's what we should be aiming for I think)

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+      });
+    }
+
+    return successResult(sessionId);
+  }
+
+  async readSessionDetails(
+    subjectIdentifier: string,
+  ): Promise<Result<SessionDetails | null>> {
+    const currentTimeInSeconds = Math.floor(Date.now() / 1000);
```
sandymay-dd, line 77: We've done this twice in this function - how about a small function within this file?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+      return errorResult({
+        errorMessage: "Session is malformed",
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+
+    return successResult(sessionId);
+  }
+
+  async readSessionDetails(
```
sandymay-dd, line 74: It feels like there's slightly too much duplication here? It feels like we should be combining this function with the readSessionId and the calling function can extract which bits of data it needs.

Thinking of options:

```
const getSessionFromSubjectIdentifier = async (subjectIdentifier: string): Promise<Session> => {

  const queryCommandInput = this.buildQueryCommandInput(subjectIdentifier)
  return await this.sendQueryCommand(queryCommandInput)
}

-------
SessionService.ts

I've missed out all the error handling here...

const getActiveSessionId = async (subjectIdentifier): Promise<string> => {
  const session = await getSessionFromSubjectIdentifier(subjectIdentifier)
  return session.sessionId
}

const getActiveSession = async (subjectIdentifier: string): Promise<Session> => {
    const session = await getSessionFromSubjectIdentifier(subjectIdentifier)
    return session
}

return 




```

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+    const sessionId = record.sessionId?.S;
+    const state = record.state?.S;
+    if (!sessionId || !state) {
+      return errorResult({
+        errorMessage: "Session is malformed",
+        errorCategory: "SERVER_ERROR",
+      });
+    }
+    const redirectUri: string | undefined = record.redirectUri?.S;
+    const sessionDetails: SessionDetails = {
```
sandymay-dd, line 112: minor one - I think this could just be Session?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const queryResponse = await this.query(input);
+    if (queryResponse.isError || queryResponse.value === null) {
+      return queryResponse;
+    }
+
+    const record: DynamoDbRecord = queryResponse.value;
+    const sessionId = record.sessionId?.S;
+    const state = record.state?.S;
+    if (!sessionId || !state) {
```
sandymay-dd, line 105: How do you feel extracting this into a function?

const isValidSession = (record: someType): boolean => !!(record.state && record.sessionId);

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -18,7 +18,7 @@ export async function lambdaHandlerConstructor(
     logger.log("ENVIRONMENT_VARIABLE_MISSING", {
       errorMessage: configResult.value.errorMessage,
     });
-    return serverError500Response;
+    return serverErrorResponse;
```
sandymay-dd, line 21: thank you

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -55,20 +56,37 @@ export async function lambdaHandlerConstructor(
     logger.log("INTERNAL_SERVER_ERROR", {
       errorMessage: getSubFromTokenResult.value.errorMessage,
     });
-    return serverError500Response;
+    return serverErrorResponse;
+  }
+
+  const sessionService = dependencies.sessionService(config.SESSION_TABLE_NAME);
+  const getActiveSessionDetailsResult =
```
sandymay-dd, line 63: minor (as above) - I'm not keen on 'details' in the name

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -146,7 +146,7 @@ export async function lambdaHandlerConstructor(
   // Create a session
   const sessionService = dependencies.sessionService(config.SESSION_TABLE_NAME);
 
-  const activeSessionResult = await sessionService.getActiveSession(
+  const activeSessionResult = await sessionService.getActiveSessionId(
```
sandymay-dd, line 149:
```suggestion
  const activeSessionIdResult = await sessionService.getActiveSessionId(
```

backend-api/src/functions/services/sessionService/sessionService.ts
```diff
@@ -0,0 +1,58 @@
...
+    subjectIdentifier: string,
+  ): Promise<Result<string | null>> {
+    return await this.dynamoDbAdapter.readSessionId(subjectIdentifier);
+  }
+
+  async getActiveSessionDetails(
+    subjectIdentifier: string,
+  ): Promise<Result<SessionDetails | null>> {
+    return await this.dynamoDbAdapter.readSessionDetails(subjectIdentifier);
+  }
```
sandymay-dd, line 29: This is looking so clean Ana!

sandymay-dd: Robert Martin would be proud

backend-api/src/functions/services/sessionService/tests/sessionService.test.ts
```diff
@@ -0,0 +1,300 @@
...
+        expect(result.value).toStrictEqual({
+          sessionId: "mockSessionId",
+          state: "mockState",
+          redirectUri: "redirectUri",
+        });
+      });
+    });
+  });
+
+  describe("Create sa ession", () => {
```
sandymay-dd, line 212: 🍺

backend-api/src/functions/services/sessionService/sessionService.ts
sandymay-dd: It's going to be challenging keeping this as clean once we have the the try catch blocks when we throw the error!

maybe 

```
try {
const session = await.this.dynamoDbAdapter.getSessionBySubjectIdentifier(subjectIdentifier)
// validate session response if needs be
catch(error) {
  return generateResultFromError (error)
}


```

backend-api/template.yaml
```diff
@@ -857,6 +857,27 @@ Resources:
...
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - dynamodb:Query
+                Resource:
+                  - Fn::GetAtt: [ "SessionsTable", "Arn" ]
+                  - !Sub ${SessionsTable.Arn}/index/subjectIdentifier-timeToLive-index
+                Condition:
+                  ForAllValues:StringEquals:
```
sandymay-dd, line 871: The list of attributes here is different from the list in the AsyncCredentialLambdaRole (?)

ahickmann-gds: Yes because this lambda also needs state and redirectUri?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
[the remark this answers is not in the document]
ahickmann-gds: I'd say the service is called DynamoDB so I don't like the idea of calling it "dynamo". 
Also, we already have a Session Service wrapping around this to make it very clear that this is about getting and creating sessions. We don't call this class directly anywhere apart from within the Session Service, I think.

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: I did not change the part of the original query that was meant to be doing that (i.e. retrieving the most recent session). I didn't test it because I expected that if that code was merged, it was tested for this scenario and I could just re-use it and adapt the query to return the session values I wanted.

I think that setting `ScanIndexForward` to `false` means that it will return the most recent added session (this is part of the code that's in `main`): 

`If ScanIndexForward is true, DynamoDB returns the results in the order in which they are stored (by sort key value). This is the default behavior. If ScanIndexForward is false, DynamoDB reads the results in reverse order by sort key value, and then returns the results to the client.
`

However, I can do a manual test as well...

sandymay-dd: I'm just thinking about this more... as we've updated the POST /credential endpoint to only create a session if one isn't available, I think we're fine here.

I agree with your analysis in any case

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Done but I did not put it in a separate file - it's within DynamoDB adapter. Perhaps it should be in SessionService?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Done

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: To be discussed again and addressed in a new PR if necessary

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: I like that idea but I would apply it if we had more attributes to validate? 

I've now removed the need to check for `.S` as I'm using the `unmarshall` method in the adapter...

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Changed to Session since we have either sessionId or session details, so the latter can just be session :D

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: I know :(

ahickmann-gds: Is it better now? @sandymay-dd

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Method no longer exists :D

ahickmann-gds: Method no longer exists

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Fixed and method no longer exists

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
ahickmann-gds: Done

backend-api/src/functions/services/sessionService/sessionService.ts
ahickmann-gds: Can we come back to this if necessary?

backend-api/template.yaml
ahickmann-gds: Correction: removing `state` as we need `sessionState` instead

backend-api/src/functions/asyncCredential/sessionService/dynamoDbClient.ts
ahickmann-gds: Done!

backend-api/src/functions/services/sessionService/tests/sessionService.test.ts
ahickmann-gds: Fixed...

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Done...

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
ahickmann-gds: Renamed to getActiveSessionIdResult

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const queryCommandOutput: QueryCommandOutput =
+      await this.dynamoDbClient.send(new QueryCommand(input));
+
+    const items = queryCommandOutput.Items;
+    if (!items || items.length === 0) {
+      return null;
+    }
+
+    return unmarshall(items[0]);
```
jmooney-dd, line 65: This is cool I had not heard of this before!

ahickmann-gds: I want to use `marshall()` when we create the session but that will be done in a separate PR as it requires a bit of testing to ensure all is the same

backend-api/template.yaml
```diff
@@ -703,6 +701,8 @@ Resources:
           Projection:
             NonKeyAttributes:
               - sessionState
+              - state
```
jmooney-dd, line 704: Just re the conversation we had at the BE catchup this morn, do we kneed both sessionState and state in here? I'm happy that you know what this answer should be so if these values are correct please dismiss

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export type DatabaseRecord = Record<string, NativeAttributeValue>;
+
+export class DynamoDbAdapter {
+  private readonly tableName: string;
+  private readonly dynamoDbClient = new DynamoDBClient({
+    region: process.env.REGION,
+    maxAttempts: 2,
+    requestHandler: new NodeHttpHandler({
+      connectionTimeout: 29000,
```
jmooney-dd, line 26: Should these be changed to the max timeout of the lambda that uses them? Or do we not care? I personally don't really have much thought on it, I just bring it up as we changed these when we were updating lambda time outs recently

backend-api/template.yaml
ahickmann-gds: We don't! I removed state from one place in the template but forgot about it here. Thanks!

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Yes, I completely forgot about this one as I merged a PR the other day to change all timeouts to match the lambda's. Thank you!

jmooney-dd:
Have we run the existing API tests against these changes?

ahickmann-gds:
> Have we run the existing API tests against these changes?

We have not (because I am still deploying my stacks after deleting them yesterday) but I will not merge this PR without re-testing all scenarios for both credential and active session.

backend-api/template.yaml
```diff
@@ -703,6 +701,7 @@ Resources:
           Projection:
             NonKeyAttributes:
               - sessionState
+              - redirectUri
```
jmooney-dd, line 704: Have you deployed from main and then redeployed with these changes? I am wondering whether adding this will make the deployment fail? If it does then we might have to do that 2 step dance with the sessions table

ahickmann-gds: I re-deployed my backend stack a few minutes ago and re-tested so I am not sure what the issue is?

jmooney-dd: Was the stack you deployed fresh? I'm wanting to test a stack that previously just had sessionState as a NonKeyAttribute in it's sessionTable and then update that stack with an updated sessions table with the new NonKeyAttribute to see if that will cause a deployment fail. 

I ask because it happened to me yesterday when I was playing around with Dynamo

backend-api/src/functions/adapters/dynamoDbAdapter.ts
sandymay-dd: hmmm I'm torn

sandymay-dd: If we update the DynamoDBAdapter file name to include the word Session I'd love for it to be in that file

sandymay-dd: Let's keep it in the DynamoDB file as that's the only file that I think should know about this. If we find in other lambdas that they need access to the authSessionState we can relocate!

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const queryCommandOutput: QueryCommandOutput =
+      await this.dynamoDbClient.send(new QueryCommand(input));
+
+    const items = queryCommandOutput.Items;
+    if (!items || items.length === 0) {
+      return null;
+    }
+
+    return unmarshall(items[0]);
```
sandymay-dd, line 65: I love this - never new about this. Thanks @ahickmann-gds !!

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -0,0 +1,118 @@
...
+        throw error;
+      }
+    }
+  }
+
+  private getTimeNowInSeconds() {
+    return Math.floor(Date.now() / 1000);
+  }
+
+  formatAttributesToGet(attributes: string[]): string {
```
sandymay-dd, line 115: please may we move this up to where it's used so it's a bit easier to look at the usage/implementation at the same time

sandymay-dd: could this be a private function?

sandymay-dd: As soon as I looked at this function alongside where it's used, the naming became crystal clear

ahickmann-gds: I wanted to have all helpers/private methods at the bottom 😞  - you will need to "find" this method within the file anyway

ahickmann-gds: I thought it was private!!!! Sorry :(

backend-api/src/functions/services/sessionService/sessionService.ts
sandymay-dd, line 1: interesting one to think about - if 'sessions' is in the name of the parent folder, do we need the sub folder to contain 'session' as well 👀

backend-api/src/functions/adapters/dynamoDbAdapter.ts
ahickmann-gds: Maybe the name should be formatAsProjectionExpression() ... @sandymay-dd ?

ahickmann-gds: I think that's much better...

backend-api/src/functions/asyncActiveSession/configService/configService.ts
```diff
@@ -14,7 +15,12 @@ export class ConfigService implements IGetConfig<Config> {
         errorCategory: "SERVER_ERROR",
       });
     }
-
+    if (!env.SESSION_TABLE_NAME) {
```
sandymay-dd, line 18: feels like there's a bit of repetition in this file that we could avoid - not for now, but 🥘 for 🤔

backend-api/template.yaml
jmooney-dd: Currently testing

ahickmann-gds: Thank you so much @jmooney-dd ⭐

backend-api/src/functions/services/sessionService/sessionService.ts
ahickmann-gds: `services/session/sessionService.ts` - how does that sound @sandymay-dd?

sandymay-dd: will we have multiple services per business domain do you think?

sandymay-dd: could we get away with for now:

services/session.ts

sandymay-dd: I'll let you self resolve this one, it's not one I'm going to die on a hill for - it's so minor and I could be swayed with any solution we've proposed so far (even the original)

backend-api/template.yaml
jmooney-dd: Have tested and the deployment failed.

What I did:
- Deployed a `backend-api` and `sts-mock` from main. This stack only has ONE non key attribute listed
- Using the same stack, I then tried to deploy the changes in this PR which has a second non key attribute listed, but it failed

Resource that failed to deploy: `SessionsTable`
Error:
Resource handler returned message: "Cannot update GSI's properties other than Provisioned Throughput and Contributor Insights Specification. You can create a new GSI with a different name." (RequestToken: 6948247f-d788-a7c7-0869-dd4ae45270c6, HandlerErrorCode: InvalidRequest)

Possible solution
You could rename the SessionsTable in this PR to effectively create a new one destroying the old (not a live service so this is fine). You could then raise a follow up PR to change the name back.

```
    Properties:
      TableName: !Sub ${AWS::StackName}-sessions-table-${Environment}-2
```

@anahickmann

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.test.ts
```diff
@@ -263,28 +263,87 @@ describe("Async Active Session", () => {
...
+            event,
+          );
+
+          expect(result).toStrictEqual({
+            headers: { "Content-Type": "application/json" },
+            statusCode: 200,
+            body: JSON.stringify({
+              sessionId: "mockSessionId",
+              redirectUri: "https://mockUrl.com/redirect",
+              state: "mockState",
```
jmooney-dd, line 344: Question, is the state value returned here the `sessionState` or `state` from the session? If `sessionState` we could have the value here `mockSessionState` to make it a bit more obvious for people where that value you came from?

If it comes from `state` rather than `sessionState` then no issue with this

backend-api/src/functions/asyncCredential/asyncCredentialHandler.test.ts
```diff
@@ -829,114 +894,32 @@ describe("Async Credential", () => {
...
+            new MockSessionServiceCreateSuccessResult();
 
           const result = await lambdaHandlerConstructor(dependencies, event);
 
+          expect(mockEventService.auditEvents[0]).toEqual(
+            "DCMAW_ASYNC_CRI_START",
+          );
           expect(mockLogger.getLogMessages()[0].logMessage.message).toBe(
-            "ERROR_CREATING_SESSION",
+            "COMPLETED",
```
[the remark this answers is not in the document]
ahickmann-gds: I am aware that we don't have a STARTED log. I made the decision to not change this in this PR because I did not want to update every single unit test.

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.test.ts
ahickmann-gds: Mock response has been fixed

backend-api/template.yaml
ahickmann-gds: Thanks @jmooney-dd. When I deployed a test stack to test my code changes I had to change the table name to make it work as well. I will do the same when I merge.

jmooney-dd (approved):

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -55,20 +56,37 @@ export async function lambdaHandlerConstructor(
...
+    return notFoundResponse;
   }
 
   logger.log("COMPLETED");
 
   return {
+    headers: { "Content-Type": "application/json" },
     statusCode: 200,
-    body: "Hello, World",
+    body: JSON.stringify(session),
```
jmooney-dd, line 83: If we're returning session, which now includes sessionState rather than session as a property, should we update the api spec?

ahickmann-gds: I am not going to update the API spec now - we need to properly review that spec as there are other things that need to be checked

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        expect(result.value).toStrictEqual({
+          errorMessage: "Error getting session ID - Error: Mock DB Error",
+          errorCategory: "SERVER_ERROR",
+        });
+        expect(result.isError).toBe(true);
+      });
+    });
+
+    describe("Given the query response does not contain the Items field", () => {
```
sandymay-dd, line 45: Is this a server error? Trying to work out what this means technically - sorry!

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+  beforeEach(() => {
+    sessionService = new SessionService("mockTableName");
+    dynamoDbMockClient = mockClient(DynamoDBClient);
+  });
+
+  beforeAll(() => {
+    jest.useFakeTimers().setSystemTime(new Date("2024-03-10"));
+  });
+
+  describe("Get active session ID", () => {
```
sandymay-dd, line 30: we should explicitly test the scenario where the query returns two sessions I think

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        expect(result.value).toStrictEqual({
+          errorMessage: "Error getting session - Error: Mock DB Error",
+          errorCategory: "SERVER_ERROR",
+        });
+        expect(result.isError).toBe(true);
+      });
+    });
+
+    describe("Given the query response does not contain the Items field", () => {
```
sandymay-dd, line 116: same here as above

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        const result = await sessionService.getActiveSessionId("mockSub");
+
+        expect(result.isError).toBe(false);
+        expect(result.value).toEqual("mockSessionId");
+      });
+    });
+  });
+
+  describe("Get active session details", () => {
```
sandymay-dd, line 101: same on explicitly testing when two items are returned.

I'm starting to think this should be an error - as it implies we've let two users into the system when POST /credential should not create a new record if one already exists?

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+
+        expect(result.isError).toBe(true);
+        expect(result.value).toEqual({
+          errorCategory: "SERVER_ERROR",
+          errorMessage: "Session is malformed",
+        });
+      });
+    });
+
+    describe("Given the item returned is missing the attribute clientState", () => {
```
sandymay-dd, line 154: I think we might be missing this test in the block above

backend-api/src/functions/services/session/tests/sessionService.test.ts
ahickmann-gds: It won't return two sessions because the query includes `Limit: 1`

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -0,0 +1,304 @@
...
+        expect(result.value).toStrictEqual({
+          errorMessage: "Error creating session - Error: Mock DB Error",
+          errorCategory: "SERVER_ERROR",
+        });
+      });
+    });
+
+    describe("Given a session with the same session ID already exists", () => {
+      it("Returns error response", async () => {
+        const mockError = new ConditionalCheckFailedException({
```
sandymay-dd, line 240: I don't know why I keep on loving this, but I do

backend-api/src/functions/services/session/tests/sessionService.test.ts
sandymay-dd: Thank you, yes. 

I supposed I was thinking about the fact we do items[0]. if we changed it to Limit:2 we wouldn't have a test that fails.

But I'm happy with what we've got

sandymay-dd: And I assume this is why you want jest mocks!

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
sandymay-dd, line 83: Please may we explicitly have this as a full object? Reasons for asking (given this functionality is already good):
- I'd like to be able to see when looking at the handler code very clearly what response we're giving
- it makes me nervous that we're dynamically tying this response to the behaviour of a method in a different service? I know our type system/mocks would help us catch this, but I'd rather be declarative if poss

sandymay-dd (approved):

backend-api/src/functions/services/session/tests/sessionService.test.ts
ahickmann-gds: I think we are looking at these tests again simply because the code has been touched - they are pretty much the same in main and I may just have renamed them.

I don't think it's a server error. DynamoDB is not throwing an exception. 

`Items` is an array that may be empty if nothing was found; however, there is this: Screenshot 2024-10-11 at 15 37 10

The type indicates that `Items` could be `undefined` so we need to check that it is not undefined in order to proceed. I don't know why this potential undefined exists and I don't know what it means. I have looked into this briefly in the past and didn't find anything obvious. I think the official documentation is very explicit about Items being an array so I don't think it will ever be undefined, but I don't know...

backend-api/src/functions/services/session/tests/sessionService.test.ts
ahickmann-gds: It does the job, I guess :)

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
ahickmann-gds: I don't agree with that. This would mean creating exactly the same object we are creating in `SessionService` and then doing exactly the same checks (is redirect_uri defined or not) in the handler. We will basically write the same code twice. We need to be able to trust our code and trust that people aren't making silly changes that will break everything. And our unit tests and API tests make sure that this doesn't happen.

ahickmann-gds: If we are going to do that then we might as well just move the entire logic to the handler

sandymay-dd: fine to resolve this - thanks for the viewpoint

backend-api/src/functions/services/session/tests/sessionService.test.ts
sandymay-dd: I find sometimes the AWS optional parameters in the SDK responses very unhelpful and unclear.

sandymay-dd: I think we are looking at these tests again simply because the code has been touched - they are pretty much the same in main and I may just have renamed them.

Yep agreed - it's a large refactor so everything sort of comes under scrutiny!

Good to resolve, and thank you for digging into the schema.

backend-api/src/functions/services/session/tests/sessionService.test.ts
sandymay-dd: Right you are.

backend-api/src/functions/services/session/tests/sessionService.test.ts
sandymay-dd: as above-  resolved.

sandymay-dd (approved):

PR #198, opened 2024-10-04
DCMAW-8389: test private and proxy API in the pipeline

ahickmann-gds:
​DCMAW-8389

## Checklists

### DIFF
--- a/backend-api/run-tests.sh
+++ b/backend-api/run-tests.sh
@@ -7,6 +7,7 @@ remove_quotes() {
   echo "$1" | tr -d '"'
 }
 export PROXY_API_URL=$(remove_quotes "$CFN_ProxyApiUrl")
+export PRIVATE_API_URL=$(remove_quotes "$CFN_PrivateApiUrl")
 export SESSIONS_API_URL=$(remove_quotes "$CFN_SessionsApiUrl")
 
 mkdir -pv results
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -1374,3 +1374,7 @@ Outputs:
     Condition: ProxyApiDeployment
     Description: Proxy API Gateway Pretty DNS
     Value: !Sub https://${ProxyApiDomainName}
+
+  PrivateApiUrl:
+    Description: Private API Gateway base URL
+    Value: !Sub https://${PrivateApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/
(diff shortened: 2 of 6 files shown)


backend-api/template.yaml
```diff
@@ -1373,4 +1373,8 @@ Outputs:
   ProxyApiUrl:
     Condition: ProxyApiDeployment
     Description: Proxy API Gateway Pretty DNS
-    Value: !Sub https://${ProxyApiDomainName}
\ No newline at end of file
+    Value: !Sub https://${ProxyApiDomainName}
+
+  PrivateApiUrl:
+    Description: Private API Gateway base URL
+    Value: !Sub https://${PrivateApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}/
```
sandymay-dd, line 1380: This has a `/` is this intentional? I don't think the proxyApiUrl has one so the tests might fail?

backend-api/run-tests.sh
```diff
@@ -7,6 +7,7 @@ remove_quotes() {
   echo "$1" | tr -d '"'
 }
 export PROXY_API_URL=$(remove_quotes "$CFN_ProxyApiUrl")
+export PRIVATE_API_URL=$(remove_quotes "$CFN_PrivateApiUrl")
```
sandymay-dd, line 10: I really want to get away from calling this Private API url in the not to distant future. Not for this PR. Private describes its implementation rather than it's purpose.

backend-api/tests/api-tests/utils/apiInstance.ts
```diff
@@ -9,15 +11,51 @@ function getInstance(baseUrl: string) {
...
+
+  apiInstance.interceptors.request.use(interceptor);
 
   return apiInstance;
 }
 
 function getSessionsApiInstance() {
-  const sessionsApiUrl = process.env.SESSIONS_API_URL;
-  if (!sessionsApiUrl)
+  const apiUrl = process.env.SESSIONS_API_URL;
```
sandymay-dd, line 39: what's the rational for this change ?

sandymay-dd (dismissed):

backend-api/tests/api-tests/utils/apiInstance.ts
ahickmann-gds: Why make the `const` name longer? It's inside `getSessionsApiInstance()` so to me it's obvious it's the session API URL. It could even be just `url` to simplify...

sandymay-dd (approved):

PR #205, opened 2024-10-07
DCMAW-8389: use aws4Interceptor with proxy API only

ahickmann-gds:
​DCMAW-8389

### What changed
- Fix API tests so that `aws4Interceptor` is only used with the proxy API and not the others

Screenshot 2024-10-07 at 16 35 47

### Why did it change
- `aws4Interceptor` not required for public and private APIs 

## Checklists

### DIFF
--- a/backend-api/tests/api-tests/utils/apiInstance.ts
+++ b/backend-api/tests/api-tests/utils/apiInstance.ts
@@ -4,7 +4,7 @@ import "dotenv/config";
 import { aws4Interceptor } from "aws4-axios";
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers";
 
-function getInstance(baseUrl: string, isProxy: boolean = false) {
+function getInstance(baseUrl: string, isProxyApi: boolean = false) {
   const apiInstance = axios.create({ baseURL: baseUrl });
   axiosRetry(apiInstance, {
     retries: 2,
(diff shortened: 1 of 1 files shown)


backend-api/tests/api-tests/utils/apiInstance.ts
```diff
@@ -4,33 +4,30 @@ import "dotenv/config";
 import { aws4Interceptor } from "aws4-axios";
 import { fromNodeProviderChain } from "@aws-sdk/credential-providers";
 
-function getInstance(baseUrl: string) {
+function getInstance(baseUrl: string, isProxyApi: boolean = false) {
```
sandymay-dd, line 7: I think this might just be personal preference, but I don't think the getInstance function should know about the ProxyApi. 

What about...

```suggestion
function getInstance(baseUrl: string, useAwsSigv4Signing: boolean = false) {
```

sandymay-dd (approved):

backend-api/tests/api-tests/utils/apiInstance.ts
sandymay-dd: I'd also be keen on the boolean (and the baseUrl if you desire) going into an object. 

I'd much rather see...

`const apiInstance = getInstance({ baseUrl: 'https://something.com', useAwsSigv4Signing: true}`

sandymay-dd (approved):

PR #211, opened 2024-10-09
DCMAW-8229: update Lambda timeouts and AWS clients

ahickmann-gds:
​DCMAW-8229

### What changed
- Pass an object to `KMSClient` instead of an array of objects
- Decrease AWS client timeouts from 29 seconds to 5 seconds
- Delete `src/functions/services/clientRegistryService/ssmClient.ts` as not in use
- Remove `console.log()` method in `src/functions/services/events/eventService.ts`
- Increase Lambda timeout to 5 seconds in the backend and STS mock SAM templates

**API tests** 
Screenshot 2024-10-09 at 17 40 39

### Why did it change
- Lambda timeout is currently set to 3 seconds and some functions have been timing out
- If Lambda timeout is set to 5 seconds, AWS clients timeouts should not be 29 seconds

## Checklists


backend-api/src/functions/services/events/sqsClient.ts
sandymay-dd, line 1: side note - the location of all our AWS clients seems a bit sporadic! Very out of scope of this

PR #215, opened 2024-10-11
[DCMAW-10280] Remove test stage from Staging pipeline

mthkbarber:
​DCMAW-10280

### What changed
Remove the TestImageUriRepository value from the Staging pipeline parameters as we do not run the test container in the Staging pipeline.

Update sessions api rate limits to allow tests to run on sessions api

### Why did it change
We don't run tests in Staging.

Test stage removed from pipeline:

image

### DIFF
--- a/infra/terraform/secure_pipelines/mob_async_backend_pl.tf
+++ b/infra/terraform/secure_pipelines/mob_async_backend_pl.tf
@@ -71,8 +71,7 @@ locals {
 
       SigningProfileArn        = one(data.aws_cloudformation_stack.signer_build[*].outputs["SigningProfileArn"])
       SigningProfileVersionArn = one(data.aws_cloudformation_stack.signer_build[*].outputs["SigningProfileVersionArn"])
-
-      TestImageRepositoryUri = one(data.aws_cloudformation_stack.mob_async_backend_tir_build[*].outputs["TestRunnerImageEcrRepositoryUri"])
+      
     }
 
     integration = {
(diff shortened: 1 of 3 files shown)


infra/terraform/secure_pipelines/mob_async_backend_pl.tf
```diff
@@ -71,8 +71,7 @@ locals {
 
       SigningProfileArn        = one(data.aws_cloudformation_stack.signer_build[*].outputs["SigningProfileArn"])
       SigningProfileVersionArn = one(data.aws_cloudformation_stack.signer_build[*].outputs["SigningProfileVersionArn"])
-
-      TestImageRepositoryUri = one(data.aws_cloudformation_stack.mob_async_backend_tir_build[*].outputs["TestRunnerImageEcrRepositoryUri"])
+
```
sandymay-dd, line 74: Is there any issue here that these locals have _build in the name?

sandymay-dd: (and the ones in 66,67)

mthkbarber: Its because its how they are named [here](https://github.com/govuk-one-login/mobile-id-check-async/blob/main/infra/terraform/secure_pipelines/main.tf). I think its how Terraform is able to access CFN stack outputs.

sandymay-dd (dismissed):

PR #216, opened 2024-10-11
[DCMAW-10280]Add temporary values for STS jwks URL in higher environments

mthkbarber:
​DCMAW-10280

### What changed
EnvironmentVariable mappings have been uncommented but the values remain as empty strings.

Add test to ensure EnvironmentVariable values are not inadvertently changed.

### Why did it change
To fix Staging deployment

Test passes when EnvironmentalVariables are correct:

image

Test to ensure tests fail when values are changed:

image


sandymay-dd:
Please may we update the PR title so it's easy to look back on main on git log: perhaps - Add temporary values for STS jwks URL in higher environments

sandymay-dd (dismissed):

sandymay-dd (approved):

PR #217, opened 2024-10-11
DCMAW-8229: Updates stack deployment guidance for backend-api and STS mock

jmooney-dd:
​DCMAW-8229

## Checklists


backend-api/README.md
sandymay-dd, line 1: Whilst I applaud this. And potentially we should look at getting it in tactically.

I'd _much_ rather script this so it's plug and play.

backend-api/README.md
```diff
@@ -49,6 +49,38 @@ sh generate_env_file.sh <stack_name>
...
+export backend-stack-name=<stack-name>
+```
+
+### Deploy a custom STS Mock stack
+
+Follow the instructions in STS Mock [README](../sts-mock/README.md) to deploy the custom STS Mock stack.
+
+### Update Environment Variables for Backend API
+
+In the backend-api [template](template.yaml), update the environment variable `STSJWKSENDPOINT` for the dev environment with the following value:
```
sandymay-dd, line 66: Do you think we should use a parameter override?

sts-mock/README.md
```diff
@@ -57,6 +57,20 @@ npm run lint
...
+Update `ASYNC_BACKEND_BASE_URL` environment variable in the STS Mock [template](../template.yaml). The new value for this URL should be:
+
+```bash
+https://sessions-<backend-api-stack-name>.review-b-async.dev.account.gov.uk
+```
+
+Please note:
+
+* You can still update this value even if you have not actually deployed a `backend-api` custom stack yet
+* You can skip this step if you are deploying/testing the STS mock in isolation
```
sandymay-dd, line 72: Is this true? The STS mock calls the backend well-known endpoint 👀

jmooney-dd: By skip I mean skip updating the variable, by default it will call the main backend-api well known endpoint, I.e. it doesn't need to call a custom be-api to test in isolation. Or is this also not true?

sandymay-dd: Yep it should be fine

PR #228, opened 2024-10-16
DCMAW-8229: validate service token

ahickmann-gds:
​DCMAW-8229

### What changed

**Log error messages:**
* Change `AUTHENTICATION_HEADER_INVALID` to `INVALID_AUTHENTICATION_HEADER`
* Add `SERVICE_TOKEN_VALIDATION_ERROR`

**Active Session handler:**
* Return 400 when JWE can't be decrypted due to a client error 
* Return 500 when JWE can't be decrypted due to a server error 
* Return 400 "Failed to validate service token" when token claims are invalid or signature is invalid
* Return 500 if token can't be validated due to a server error (e.g. we could not get the public key with which to validate the signature)
* Remove the token dependencies from the handler dependencies as they are no longer required

**Config Service:**
* Add `AUDIENCE` env var and make it equal to the sessions API domain name
* Replace the `STS_JWKS_ENDPOINT` env var with `STS_BASE_URL` env var 
* Create `STS_JWKS_ENDPOINT` in the Config Service

**JWE Decryption:**
* Return `SERVER_ERROR` when CEK can't be decrypted with KMS
* Make `decrypt()` in `SymmetricDecrypter` synchronous

**Header validation:**
* Move logic to validate JWT header and extract the `kid`- moved from the `PublicKeyGetter` to the `TokenVerifier` - this seemed to a be a better place for this bit of code and it made it easier to return the appropriate error response to the user

**Token Service:**
* Responsible for validating the header, payload and signature of token

**Token Verifier:**
* Expanded it to fetch the public key via the PublicKeyGetter

**Send HTTP Request:**
* Replaced the error response object with exceptions

**200 Valid service token and session found**
Screenshot 2024-10-16 at 17 15 25

**400 Failed to validate service token** (could be that payload or signature validation failed - a message is logged so we can tell the difference)
Screenshot 2024-10-16 at 16 54 05

**400 Failed to decrypt service token**
Screenshot 2024-10-16 at 16 54 55

**401 Invalid Authorization header**
Screenshot 2024-10-16 at 16 54 29

**404 Session not found**
Screenshot 2024-10-16 at 17 04 23

## Checklists


backend-api/src/functions/asyncActiveSession/configService/configService.ts
```diff
@@ -21,25 +23,41 @@ export class ConfigService implements IGetConfig<Config> {
...
         errorCategory: "SERVER_ERROR",
       });
     }
 
     return successResult({
       ENCRYPTION_KEY_ARN: env.ENCRYPTION_KEY_ARN,
       SESSION_TABLE_NAME: env.SESSION_TABLE_NAME,
-      STS_JWKS_ENDPOINT: env.STS_JWKS_ENDPOINT,
+      STS_BASE_URL: env.STS_BASE_URL,
+      STS_JWKS_ENDPOINT: env.STS_BASE_URL + "/.well-known/jwks.json",
```
ahickmann-gds, line 59: Not sure if we are ok with this?

backend-api/src/functions/testUtils/mockJwtBuilder.ts
```diff
@@ -3,11 +3,11 @@ import jose from "node-jose";
 
 export class MockJWTBuilder {
```
ahickmann-gds, line 4: Once this PR is merged I need to go back to this `MockJWTBuilder` and change it a bit. I didn't want to break the credential tests even more so I tried not to change it much.

backend-api/src/functions/services/http/sendHttpRequest.ts
ahickmann-gds, line 1: I believe I haven't changed the logic here - I just replaced the errorResult with an exception.

backend-api/src/functions/asyncCredential/tokenService/tests/tokenService.test.ts
ahickmann-gds, line 1: Had to update this test after I fixed the audience value in the MockJWTBuilder()

backend-api/src/functions/asyncActiveSession/jwe/jweDecrypter.ts
```diff
@@ -57,13 +57,13 @@ export class JweDecrypter implements IDecryptJwe {
     } catch (error) {
       return errorResult({
         errorMessage: `Error decrypting JWE: Unable to decrypt encryption key: ${error}`,
-        errorCategory: "CLIENT_ERROR",
+        errorCategory: "SERVER_ERROR",
```
ahickmann-gds, line 60: We need to look into this. Right now we are saying that ALL KMS decryption errors are server errors, and I think we might also want to return a client error depending on the exception type. That's something for the next ticket.

backend-api/src/functions/asyncActiveSession/jwe/symmetricDecrypter.ts
```diff
@@ -1,13 +1,13 @@
 import { createDecipheriv } from "node:crypto";
 
 export class SymmetricDecrypter implements IDecryptSymmetric {
-  async decrypt(
+  decrypt(
```
ahickmann-gds, line 4: No need for this to be async 🥹

backend-api/src/functions/asyncActiveSession/jwe/tests/jweDecrypter.test.ts
```diff
@@ -58,7 +58,7 @@ describe("Decrypt JWE", () => {
       expect(result.isError).toBe(true);
       expect(result.value).toStrictEqual({
         errorMessage:
-          "Error decrypting JWE: Unable to decrypt payload: Some mock symmetric decryption error",
+          "Error decrypting JWE: Unable to decrypt payload: Error: Some mock symmetric decryption error",
```
ahickmann-gds, line 61: I need to format these errors a bit better but again I didn't want to do this in this PR and make it even bigger

backend-api/src/functions/asyncActiveSession/registeredLogs.ts
ahickmann-gds, line 1: Can we please check that we are happy with these messages?

backend-api/src/functions/asyncActiveSession/tokenService/publicKeyGetter.ts
```diff
@@ -1,40 +1,53 @@
...
   async getPublicKey(
-    jwks: IJwks,
-    jwt: string,
+    jwksEndpoint: string,
+    kid: string,
   ): Promise<Result<Uint8Array | KeyLike>> {
-    let decodedProtectedHeader: ProtectedHeaderParameters;
+    let jwk;
     try {
-      decodedProtectedHeader = decodeProtectedHeader(jwt);
```
ahickmann-gds, line 16: I've moved this header logic to the TokenService. It made more sense there and it facilitated returning a client error when the header can't be decoded or there is no kid claim.

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -26,55 +26,59 @@ export async function lambdaHandlerConstructor(
...
+  console.log(serviceTokenJwt);
 
-  const tokenServiceDependencies = dependencies.tokenServiceDependencies;
-  const tokenService = dependencies.tokenService(tokenServiceDependencies);
-  const getSubFromTokenResult = await tokenService.getSubFromToken(
-    config.STS_JWKS_ENDPOINT,
+  const tokenService = dependencies.tokenService(config.STS_JWKS_ENDPOINT);
+  const validateServiceTokenResult = await tokenService.validateServiceToken(
+    serviceTokenJwt,
     {
```
ahickmann-gds, line 56: I thought that since we don't intend to make these values environment variables then there is no point in passing them through all the way from the handler to the `sendHttpRequest` function, so I've moved this object to `publicKeyGetter.ts`

backend-api/src/functions/asyncActiveSession/handlerDependencies.ts
```diff
@@ -5,37 +5,21 @@ import {
...
-import {
-  ITokenService,
-  ITokenServiceDependencies,
-  TokenService,
-} from "./tokenService/tokenService";
-import { TokenVerifier } from "./tokenService/tokenVerifier";
+import { ITokenService, TokenService } from "./tokenService/tokenService";
 import { IDecryptJwe, JweDecrypter } from "./jwe/jweDecrypter";
 
-const tokenServiceDependencies: ITokenServiceDependencies = {
```
ahickmann-gds, line 17: We didn't need all these dependencies here in the end but we can have a discussion about this if people would like that.

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -26,63 +26,61 @@ export async function lambdaHandlerConstructor(
...
-    if (getSubFromTokenResult.value.errorCategory === "CLIENT_ERROR") {
-      logger.log("FAILED_TO_GET_SUB_FROM_SERVICE_TOKEN", {
-        errorMessage: getSubFromTokenResult.value.errorMessage,
+  if (validateServiceTokenResult.isError) {
+    if (validateServiceTokenResult.value.errorCategory === "CLIENT_ERROR") {
+      logger.log("SERVICE_TOKEN_VALIDATION_ERROR", {
+        errorMessage: validateServiceTokenResult.value.errorMessage,
       });
-      return badRequestResponse(getSubFromTokenResult.value.errorMessage);
+      return badRequestResponse("Failed to validate service token");
```
ahickmann-gds, line 67: I've decided that if the token claims are invalid or the signature is invalid, I will return the same response. The actual problem is logged on lines 64-65. If this isn't appropriate, please do let me know/let's have a chat.

backend-api/src/functions/asyncActiveSession/tokenService/tests/tokenService.test.ts
```diff
@@ -1,443 +1,334 @@
...
+        expect(result.isError).toBe(true);
+        expect(result.value).toStrictEqual({
+          errorMessage: "Invalid kid claim",
+          errorCategory: "CLIENT_ERROR",
+        });
+      });
+    });
   });
 
-  describe("Get Sub From Token", () => {
```
ahickmann-gds, line 54: These tests have moved - they are no longer required in the TokenService. They are now in the sendHttpRequest and the publicKeyGetter test files...

backend-api/src/functions/asyncActiveSession/configService/configService.ts
sandymay-dd: I'm good with this. Assuming that it _doesn't_ make sense to move the ./well-known/jwks.json string closer to the functionality that gets the public key

backend-api/src/functions/asyncActiveSession/jwe/symmetricDecrypter.ts
sandymay-dd: haha - I'd like an async symmetric decrypter

backend-api/src/functions/testUtils/mockJwtBuilder.ts
```diff
@@ -111,6 +114,15 @@ export class MockJWTBuilder {
   };
 }
 
+const MOCK_PRIVATE_KEY = {
```
sandymay-dd, line 117: Please may we add a comment next to this saying:
- How this was generated
- That this is only used in unit tests and not used in any AWS environment

It will get flagged on a github secret scan and the comments really help people that review them :)

backend-api/src/functions/asyncActiveSession/jwe/tests/jweDecrypter.test.ts
ahickmann-gds: Already done - resolving

backend-api/src/functions/asyncActiveSession/configService/configService.ts
ahickmann-gds: It does make sense to move it closer - it makes a lot of sense! I will do that!

sandymay-dd: not really - it felt better bringing it out of the template though!

ahickmann-gds: I have moved this to the PublicKeyGetter. 
STS_JWKS_ENDPOINT has been deleted.

backend-api/src/functions/testUtils/mockJwtBuilder.ts
ahickmann-gds: I've renamed this to MOCK_SIGNING_KEY and left a small comment to say it's only used in unit tests

sandymay-dd (approved):

PR #229, opened 2024-10-18
DCMAW-8229: check KMS error type

ahickmann-gds:
​DCMAW-8229

### What changed

**KMS Adapter**
Check the exception type thrown by the KMS Client. If of type `InvalidCiphertextException` or `IncorrectKeyException`, throw a custom exception `ClientError`, otherwise just re-throw the error caught.

Screenshot 2024-10-18 at 10 02 15

Screenshot 2024-10-18 at 10 01 51

**Mock JWT Builder**
* Delete unused methods

### Why did it change
If the client sends an invalid JWE (e.g. one where the CEK has been encrypted with the wrong key or the encryption is invalid), we currently return a server error (500) to the client, which is incorrect as these clearly error client errors (400).

## Checklists

### DIFF
--- a/backend-api/src/functions/adapters/kmsAdapter.ts
+++ b/backend-api/src/functions/adapters/kmsAdapter.ts
@@ -1,9 +1,18 @@
 import {
   DecryptCommand,
   DecryptCommandOutput,
+  IncorrectKeyException,
+  InvalidCiphertextException,
   KMSClient,
 } from "@aws-sdk/client-kms";
 import { NodeHttpHandler } from "@smithy/node-http-handler";
+
+export interface IKmsAdapter {
+  decrypt: (
+    ciphertext: Uint8Array,
+    encryptionKeyId: string,
+  ) => Promise<Uint8Array>;
+}
 
 const kmsClient = new KMSClient({
   region: "eu-west-2",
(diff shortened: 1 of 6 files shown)


backend-api/src/functions/adapters/kmsAdapter.ts
```diff
@@ -19,13 +28,24 @@ export class KMSAdapter implements IKmsAdapter {
...
+        new DecryptCommand({
+          KeyId: encryptionKeyId,
+          CiphertextBlob: encryptedData,
+          EncryptionAlgorithm: "RSAES_OAEP_SHA_256",
+        }),
+      );
+    } catch (error) {
+      if (
+        error instanceof InvalidCiphertextException ||
+        error instanceof IncorrectKeyException
```
jhumbert-dd, line 43: KMS seems to throw three 500 server errors and around eight 400 errors. Is it worth doing the reverse here and checking for server errors first?

jmooney-dd: @JacksonJ2W can I check re those errors you're referring to (I've not looked at the docs)... Do those KMS 500 and 400 codes directly map to what we would return in the response? Could we could get a bad request from a client and then give KMS something bad which could cause a KMS 500 error but we'd still want to return 400? Or vice versa.

ahickmann-gds: That would be the easiest and nicest solution, but the issue is that not all client exceptions thrown by KMS mean that our client (caller) did something wrong. For example, the `InvalidKeyUsageException` is a client exception but it would only be thrown if we (i.e. async backend) make a mistake in the decrypt request we send to KMS, so it would be wrong to return our client a 400 when we made the mistake ourselves and not them.

ahickmann-gds: This is a confusing one because in this situation we are KMS' client but we also have a client of our own that we need to respond to...

jmooney-dd: I would rather return a 4XX in the active session response when we defo know the client did something wrong (as is)

ahickmann-gds: But we are now, no? 🤔

jhumbert-dd (approved):

backend-api/src/functions/adapters/kmsAdapter.ts
jmooney-dd: @anahickmann yes sorry that's what I meant by: (as is)

jmooney-dd (approved):

backend-api/src/functions/adapters/kmsAdapter.ts
ahickmann-gds: Ah ok, sorry 🤦‍♀️

jmooney-dd: No, bad wording from me 😄

PR #230, opened 2024-10-18
DCMAW-8229: Add API tests for activeSession

jhumbert-dd:
​DCMAW-8229

### What changed
- Adds API tests for the Active Session lambda
- Adds output in the template.yaml for the STS Mock API URL
- Updates generate_env_file.sh with STS Mock API URL
- Extracts test function for getting registered client

### Why did it change
To ensure we're able to test and receive the correct response when making network requests to the activeSession lambda.

#### All API tests run against main passing:

Screenshot 2024-10-18 at 12 07 46

#### Stack outputs include STS Mock API URL:

Screenshot 2024-10-18 at 12 39 06

#### generate_env_file.sh adds `STS_MOCK_API_URL` env var with correct value:

Screenshot 2024-10-18 at 12 45 26

## Checklists


backend-api/tests/api-tests/activeSession.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      expect(response.status).toBe(404);
+      expect(response.data).toStrictEqual({
+        error: "session_not_found",
+        error_description:
+          "No active session found for the given sub identifier",
+      });
+    });
+  });
+
+  describe("Given a successful call is made to /async/activeSession", () => {
```
ahickmann-gds, line 55: Can this maybe be something like "Given the request is valid and a session is found?" or just "Given a session is found"?

backend-api/tests/api-tests/activeSession.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  const asyncTokenResponse = await PROXY_API_INSTANCE.post(
+    "/async/token",
+    "grant_type=client_credentials",
+    {
+      headers: {
+        "x-custom-auth": `Basic ${clientIdAndSecretB64}`,
+      },
+    },
+  );
+  const credentialTokenResponse = await PROXY_API_INSTANCE.post(
```
ahickmann-gds, line 101: Is this meant to be `asyncCredentialResponse`?

backend-api/template.yaml
```diff
@@ -1400,3 +1400,7 @@ Outputs:
   PrivateApiUrl:
     Description: Private API Gateway base URL
     Value: !Sub https://${PrivateApi}.execute-api.${AWS::Region}.amazonaws.com/${Environment}
+
+  StsMockApiUrl:
```
sandymay-dd, line 1404: If we know what the DNS is for each environment - do we need this, or could we assume it directly in the template?

backend-api/tests/api-tests/activeSession.test.ts
sandymay-dd: Or even "Given there's an active session for this user" 👀

backend-api/tests/api-tests/activeSession.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    subject_token_type: "urn:ietf:params:oauth:token-type:access_token",
+  });
+  const stsMockResponse = await STS_MOCK_API_INSTANCE.post(
+    "/token",
+    requestBody,
+  );
+  return stsMockResponse.data.access_token;
+}
+
+async function createSessionForSub(sub: string) {
```
sandymay-dd, line 87: I feel we should be re-using test steps from the clientCredentialsGrantFlow.ts here. I'd rather get this PR merged and consider it if we have time as this will could require a wider refactor

backend-api/tests/api-tests/utils/getRegisteredClient.ts
sandymay-dd, line 1: Appreciate this is an existing folder - but I'm really not a fan of `utils` unless it's common functionality that has no business logic. E.g. converting a `string` to a `Date` is an example of something that might form a `utils` file. What we start to see is the `utils` folder containing more and more files because we aren't sure where best to put them.

Again - I think this is one for later - I wonder if the solution here is to create a test-steps folder that does common common requests

backend-api/tests/api-tests/activeSession.test.ts
jhumbert-dd: Yeah, nice - updated

backend-api/tests/api-tests/activeSession.test.ts
jhumbert-dd: Thanks - updated

backend-api/template.yaml
jhumbert-dd: This was added so that the generate_env_file.sh can create an STS_MOCK_API_URL env var for testing main and custom stacks

backend-api/tests/api-tests/activeSession.test.ts
jhumbert-dd, line 1: These tests occasionally timeout. Running the tests again usually works but I'm not sure if it's worth using a setTimeout as it's not clear it works and for how long this should be set

backend-api/tests/api-tests/activeSession.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        "x-custom-auth": `Basic ${clientIdAndSecretB64}`,
+      },
+    },
+  );
+  const credentialTokenResponse = await PROXY_API_INSTANCE.post(
+    "/async/credential",
+    {
+      sub:
+        sub ??
+        "urn:fdc:gov.uk:2022:56P4CMsGh_02YOlWpd8PAOI-2sVlB2nsNU7mcLZYhYw=",
```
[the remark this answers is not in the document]
jhumbert-dd: Updated

ahickmann-gds (approved):
This is great!

ahickmann-gds (approved):

PR #233, opened 2024-10-21
DCMAW-10416: Automate backend-api custom stack deployment to dev

jmooney-dd:
​DCMAW-10416

### What changed
- Adds `deploy_backend.sh` in `/helper-scripts`
- Adds `backend-api` `npm` script to execute (preffered as you stay in the `/backend-api` after deploying
- Updates `README`'s for `helper-scripts`, `backend-api` and `sts-mock`
- Adds override parameters in templates which script will use to link custom `backend-api` and `sts-mock` stack
- Adds execute permissions to `publish_keys_to_s3` script

### Why did it change
In order to deploy a custom `backend-api` stack to the dev environment, a developer must manually run several commands, update various environment variables, and deploy 2 custom stacks

To streamline the process, we need a script that automates the deployment workflow. The script should allow developers to deploy the custom stack with a single command, reducing the potential for errors and making the process far more efficient.

#### API tests
All passing against `backend-api` deployed using script

#### Have asked other devs to try this
- [ ] Confirmed successful

## Checklists


helper-scripts/deploy_backend.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Check correct AWS account is active
+if [ "$(aws sts get-caller-identity --output text --query 'Account')" != "211125300205" ]; then
+    echo "This script is attempting to be run in a NON DEV account. Please verify your AWS environment configuration"
+    exit 1
+fi
+
+# Ask the user if they want to build and deploy the custom backend-api
+echo
+read -r -p "You are about to deploy a custom Backend API, do you wish to continue? [y]: " yn
```
sandymay-dd, line 17:
```suggestion
read -r -p "You are about to deploy your own backend-api, do you wish to continue? [y]: " yn
```

helper-scripts/deploy_backend.sh
```diff
@@ -0,0 +1,100 @@
...
+sam build --cached
+sam deploy \
+    --stack-name "$BACKEND_STACK_NAME" \
+    --parameter-overrides "$DEV_OVERRIDE_STS_BASE_URL=https://${STS_MOCK_STACK_NAME}.review-b-async.dev.account.gov.uk" \
+    --capabilities CAPABILITY_NAMED_IAM \
+    --resolve-s3
+
+# Ask the user about deploying sts-mock
+deploy_sts_mock=false
+while true; do
```
sandymay-dd, line 47: We could simplify this to be an if condition 🤷

helper-scripts/deploy_backend.sh
```diff
@@ -0,0 +1,100 @@
...
+cd ../backend-api || exit 1
+sam build --cached
+sam deploy \
+    --stack-name "$BACKEND_STACK_NAME" \
+    --parameter-overrides "$DEV_OVERRIDE_STS_BASE_URL=https://${STS_MOCK_STACK_NAME}.review-b-async.dev.account.gov.uk" \
+    --capabilities CAPABILITY_NAMED_IAM \
+    --resolve-s3
+
+# Ask the user about deploying sts-mock
+deploy_sts_mock=false
```
sandymay-dd, line 46: is this condition used anywhere?

helper-scripts/deploy_backend.sh
```diff
@@ -0,0 +1,122 @@
...
+                ;;
+            * )
+                echo "Invalid input. Please enter 'y' or 'n'."
+                ;;
+        esac
+    done
+fi
+
+# Ask the user if they want to generate a .env file for the custom Backend API stack
+while true; do
```
sandymay-dd, line 100: If we're adding this, I'd suggest doing it automatically after deploying the backend stack. i.e. don't ask the user

helper-scripts/deploy_backend.sh
jmooney-dd: The reason I used a while loop was to keep re asking the question if the user responds with invalid response. Can you do that with an if?

sandymay-dd: ah fair

helper-scripts/deploy_backend.sh
```diff
@@ -0,0 +1,111 @@
...
+    echo
+    read -r -p "Do you want to deploy a custom sts-mock stack? [y/n]: " yn
+
+    case "$yn" in
+        [yY] )
+            deploy_sts_mock=true
+            # Build and deploy sts-mock
+            echo "Building and deploying sts-mock stack: $STS_MOCK_STACK_NAME"
+            echo
+            cd ../sts-mock || exit 1
```
sandymay-dd, line 37: I wonder if we should do an npm install command here

jmooney-dd: For what reason? Are you seeing issues?

sandymay-dd: to ensure we have the latest packages before deploying a stack and testing. If someone does, it's a 1 second additional wait on the script. Nothing lost, but potentially something gained?

helper-scripts/deploy_backend.sh
sandymay-dd, line 1: Very happy with this for now - we can iterate over time. 

There are a few steps which require user interaction - potentially not ideal if you want to run the script and get a coffee - food for thought going forward

helper-scripts/README.md
```diff
@@ -1,12 +1,34 @@
...
+#### Execute `deploy-backend-to-dev` script
+
+```bash
+# From /backend-api
+npm run deploy-backend-to-dev <your-stack-name>
+```
+
+#### Follow the prompts
+
+1. Deploying an `sts-mock` stack. This is required the first time you deploy a `backend-api` stack. It's optional for subsequent deployments.
```
sandymay-dd, line 32: Is it technically required? I think this guidance should be in the echo statements from the script rather than the readme potentially?

helper-scripts/deploy_backend.sh
jmooney-dd: Done

helper-scripts/README.md
jmooney-dd: Active Session API tests will fail if it's not deployed so to have a full working `backend-api` I believe it is required. Do you think otherwise? I suppose you could argue if you're testing asyncToken or asyncCredential in isolation it wouldn't be required but quite niche to put in a guide.

Will see what it looks like with guidance echoed in terminal now..

sandymay-dd: I wonder if we should just always deploy both stacks here and simplify the user inputs? If it's painful, we can add the checks in?

sandymay-dd: To run the sts api tests, you need the backend. If we have environment coupling it feels sensible to deploy both at once

sandymay-dd (approved):

helper-scripts/README.md
jmooney-dd: Will iterate in future as discussed

PR #239, opened 2024-10-23
[DCMAW-10446] Vanity URL for PrivateApiGateway

dothomson:
Ticket: https://govukverify.atlassian.net/browse/DCMAW-10446

Implements a workaround for custom domain names for a private api gateway. By using a load balancer pointed at the VPCEndpoint we're able to create a record which directs traffic to the load balancer. 

The popular tutorials suggest creating a private hosted zone to hold the DNS record. This first attempt is using the established Public Hosted zone for the CNAME record to the load balancer DNS.

Tests: 

Initial tests show the proxy lambda, which is using the vanity url, is successfully working. 
image
image

And the load balancer is in use. 

image

### DIFF
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -35,18 +35,33 @@ Mappings:
     dev:
       ApiBurstLimit: 10
       ApiRateLimit: 10
+      VpceIPAddressA: 10.0.10.23
+      VpceIPAddressB: 10.0.13.96
+      VpceIPAddressC: 10.0.14.16
     build:
       ApiBurstLimit: 10
       ApiRateLimit: 10
+      VpceIPAddressA: 10.0.10.132
+      VpceIPAddressB: 10.0.13.150
+      VpceIPAddressC: 10.0.15.219
     staging:
       ApiBurstLimit: 0
       ApiRateLimit: 0
+      VpceIPAddressA: 10.0.10.12
+      VpceIPAddressB: 10.0.13.85
+      VpceIPAddressC: 10.0.15.161
     integration:
       ApiBurstLimit: 0
       ApiRateLimit: 0
+      VpceIPAddressA: 10.0.10.42
+      VpceIPAddressB: 10.0.13.207
+      VpceIPAddressC: 10.0.14.190
     production:
       ApiBurstLimit: 0
       ApiRateLimit: 0
+      VpceIPAddressA: 10.0.11.211
+      VpceIPAddressB: 10.0.13.200
+      VpceIPAddressC: 10.0.15.16
 
   SessionsApigw:
     dev:
@@ -289,6 +304,7 @@ Resources:
   PrivateApiLoadBalancer:
     Type: AWS::ElasticLoadBalancingV2::LoadBalancer
     Properties:
+      Name: !Sub private-api-${AWS::StackName}
       IpAddressType: ipv4
       Scheme: internal
       Subnets:
@@ -313,6 +329,7 @@ Resources:
   PrivateApiTargetGroup:
     Type: AWS::ElasticLoadBalancingV2::TargetGroup
     Properties:
+      Name: !Sub private-api-${AWS::StackName}
       HealthCheckEnabled: true
       HealthCheckIntervalSeconds: 10
       HealthCheckPort: 443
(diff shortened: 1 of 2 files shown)


backend-api/template.yaml
```diff
@@ -243,6 +243,95 @@ Resources:
           Parameters:
             Location: ./openApiSpecs/async-private-spec.yaml
 
+  PrivateApiDomainName:
+    Type: AWS::ApiGateway::DomainName
+    Properties:
+      DomainName: !Sub
+        - private-${AWS::StackName}.${DNS_RECORD}
```
sandymay-dd, line 250: @darrendunford - did you have any opinions on naming? 

This would come through as private-mob-async-backend.${DNS_RECORD}. I was potentially wondering about this being 'oauth-mob-async-backend.${DNS_RECORD}

backend-api/template.yaml
```diff
@@ -625,7 +714,13 @@ Resources:
     Properties:
       Environment:
         Variables:
-          PRIVATE_API_URL: !Sub https://${PrivateApi.RestApiId}.execute-api.eu-west-2.amazonaws.com/${Environment}
+          # PRIVATE_API_URL: !Sub https://${PrivateApi.RestApiId}.execute-api.eu-west-2.amazonaws.com/${Environment}
```
darrendunford, line 717: can the comment be removed now?

backend-api/template.yaml
```diff
@@ -243,6 +243,95 @@ Resources:
...
+      UnhealthyThresholdCount: 3
+      Port: 443
+      Protocol: TLS
+      Targets:
+        - Id: 10.0.10.23 # subnet-0d43f751301129abd (devplatform-vpc-PrivateSubnetA) eu-west-2c (euw2-az1)
+          Port: 443
+        - Id: 10.0.13.96 # subnet-094e66815780c4556 (devplatform-vpc-PrivateSubnetB) eu-west-2a (euw2-az2)
+          Port: 443
+        - Id: 10.0.14.16 # subnet-0d90ca1c3b3a83028 (devplatform-vpc-PrivateSubnetC) eu-west-2b (euw2-az3)
+          Port: 443
```
darrendunford, line 331: where do these hard coded values come from? are they stable?

backend-api/template.yaml
```diff
@@ -243,6 +243,95 @@ Resources:
...
+  PrivateApiTargetGroup:
+    Type: AWS::ElasticLoadBalancingV2::TargetGroup
+    Properties:
+      HealthCheckEnabled: true
+      HealthCheckIntervalSeconds: 10
+      HealthCheckPort: 443
+      HealthCheckProtocol: TCP
+      HealthCheckTimeoutSeconds: 10
+      HealthyThresholdCount: 3
+      UnhealthyThresholdCount: 3
```
darrendunford, line 322: are the targets able to service the health check?

dothomson: The targets are reporting healthy at the moment 👍  

image

backend-api/template.yaml
dothomson: The values are currently the IP addresses for the ENIs of the Api Gateway VPC Endpoint. I need to add the values for the other environments. 

https://docs.aws.amazon.com/vpc/latest/privatelink/privatelink-access-aws-services.html#aws-service-subnets-zones

> The IP addresses of an endpoint network interface will not change during the lifetime of its VPC endpoint.

We can be confident these values will persist, but they would need to be updated if we moved VPCs / recreated the devplatform VPC.

backend-api/template.yaml
darrendunford: since we're using a public hosted zone for this the DNS name is visible to the world. I guess `oauth-...` would be better as the word `private` may attract underwanted attention from those who seek to gain access to something private ???

backend-api/template.yaml
darrendunford: is there any way of getting these dynamically, perhaps by having the VPC stack write them out to CF exports or something like that?

dothomson:
AWS have released custom private dns addresses as a new feature! This workaround is no longer needed. 

https://aws.amazon.com/blogs/compute/implementing-custom-domain-names-for-private-endpoints-with-amazon-api-gateway/

PR #275, opened 2024-12-02
[DO NOT MERGE] POC for getting/updating sessions in different states.

chris-cooksley-gds:
## Rationale

For the async work we'll soon need to work with async sessions in multiple states, for which it will pay to have a pattern that's type safe and easy to extend/modify.

Some objectives:
* specific session types for each state, with the fields in the type limited to those available to a session in that state
* the ability to fetch a session in a specified state, and get back an object as above
* the ability to update a session to be in a particular new state, along with any new fields to be added, and including any conditional checks on a required initial state

A lot of this is adapted from patterns used in the STS codebase, where they've seen some success.

## Overview of new code

`getSessionWithState` attempts to get a session from dynamo that's expected to be in a particular state.
* It's overloaded - this turned out to be the easiest way to make it work type-safely for different states (TS conditionals/generics never quite worked as expected, at least when we tried it on MPT).
* It'll either give you back a Success with a session object in that state (with all and only the fields we'd expect to be on a session in this state), or a failure with a reason why
* It'll fall back to a strongly consistent read if the first attempt can't find a session in this state, as the default (cheaper) eventually consistent read may not reflect recent writes/updates
* It makes use of state-specific convertor functions to convert from the Dynamo output into a particular type/shape of session, with validation applied to make sure the expected fields are present.

`updateSession` performs a particular update operation to a pre-existing session
* the update operations are polymorphic. They implement a common interface, and can all produce the necessary dynamo expressions etc to perform the update they represent. This makes them individually easy to manage/maintain/test, while also making it simpler to test the (agnostic) update method.
* Some updates come with a conditional check on a specific current session state. Others don't, on the basis that we'll already have retrieved the session and checked its state before making the update (and don't want the extra read charge from dynamo to check again). This could also be a boolean on the updateMethod, I suppose, if we didn't want to manage it at the operation level.

## Stray comments
* I've put (placeholder) logging in just using the console, to be replaced by real logging. Couple differences to our usual logging pattern here that might be interesting to chat about (but not the focus of this):
  * I log in the adapter itself rather than propagating the error up to be logged elsewhere. The idea is to capture and log specific error details as close to the source as possible, and simplify the information that then needs to be passed around elsewhere in the system.
  * I log the attempt and success of the operation as well as just failures. I've found this to be very useful for debugging external calls - e.g. we can see that we made an attempt that never returned a success or failure, or that we got a successful result (or failure) a full 2 seconds after starting the attempt.
* I've updated our use of the Result pattern to make it easier to capture 'empty' successes/failures (when we don't need any information back aside from the success/failure of an op), and to return different error information on a failure. Again not the focus of this change, just made it easier to adapt some of the MPT functions.
* I've not covered every session state for this POC - notably ASYNC_AUTH_SESSION_ABORTED is absent


backend-api/src/functions/common/session/sessionsDemo.test.ts
```diff
@@ -0,0 +1,70 @@
+import { SessionState } from "./Session";
+import { DynamoDbAdapter } from "../../adapters/dynamoDbAdapter";
+import { randomUUID } from "crypto";
+import "dotenv/config";
+import { BiometricSessionFinished } from "./updateOperations/BiometricSessionFinished";
+import { BiometricTokenIssued } from "./updateOperations/BiometricTokenIssued";
+
+describe("Sessions", () => {
+  // Log into AWS and run this to demonstrate the functionality of the new get/update methods.
```
jmooney-dd, line 9: Are there any other pre reqs to running this test? I am currently getting the following error:

ResourceNotFoundException: Requested resource not found

Which is being caught after running:

await this.dynamoDbClient.send(new PutItemCommand(input))

backend-api/src/functions/common/session/sessionsDemo.test.ts
```diff
@@ -0,0 +1,70 @@
+import { SessionState } from "./Session";
+import { DynamoDbAdapter } from "../../adapters/dynamoDbAdapter";
+import { randomUUID } from "crypto";
+import "dotenv/config";
+import { BiometricSessionFinished } from "./updateOperations/BiometricSessionFinished";
+import { BiometricTokenIssued } from "./updateOperations/BiometricTokenIssued";
+
+describe("Sessions", () => {
```
jmooney-dd, line 8: Very clear console.logs produced here, thanks! Wish I had read line 11 a little harder before running the test a few times 😆

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -32,10 +55,83 @@ export class DynamoDbAdapter {
...
+    if (session.sessionState === SessionState.BIOMETRIC_TOKEN_ISSUED) {
+      return successResult(session);
+    }
+    if (session.sessionState === SessionState.BIOMETRIC_SESSION_FINISHED) {
+      return successResult(session);
+    }
+    if (session.sessionState === SessionState.RESULT_SENT) {
+      return successResult(session);
+    }
+    return successResult(session);
```
jmooney-dd, line 132: We say if it's one of these 3 states return a successResult(session), but if not we still return the same thing. Am I missing a reason why we do this or is this just because it's POC?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -32,10 +55,83 @@ export class DynamoDbAdapter {
...
+    sessionId: string,
+    sessionState: SessionState.RESULT_SENT,
+  ): Promise<Result<BiometricSessionFinishedSession, ReadSessionError>>;
+  async getSessionWithState(
+    sessionId: string,
+    sessionState: SessionState,
+  ): Promise<Result<Session, ReadSessionError>> {
+    let output: GetItemCommandOutput;
+    try {
+      console.log("Get session attempt"); // replace with proper logging
```
jmooney-dd, line 87: In the past we only log by exception (apart from started, completed logs), do you think here there's a good reason to change this?

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished.ts
```diff
@@ -0,0 +1,23 @@
...
+  getDynamoDbUpdateExpression() {
+    return "set biometricSessionId = :biometricSessionId, sessionState = :biometricSessionFinished";
+  }
+
+  getDynamoDbConditionExpression(): string {
+    return `sessionState in (:biometricTokenIssued)`;
+  }
+
+  getDynamoDbExpressionAttributeValues() {
+    return {
```
jmooney-dd, line 15: Not relevant to this POC but for future ref, there are marshall and unmarshall functions that deal with that silly { S: 'something' } formatting. Believe @anahickmann may have used them in the async codebase already so we could look to use them in future.

backend-api/src/functions/exampleLambda/exampleLambdaConfig.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  Config,
+  getRequiredEnvironmentVariables,
+  MissingEnvVarError,
+} from "../common/utils/environment";
+import { Result } from "../utils/result";
+
+const REQUIRED_ENVIRONMENT_VARIABLES = ["SESSIONS_TABLE"] as const;
```
jmooney-dd, line 8: What is the value in asking for the envVars you want like you do here vs having envVars defined for each lambda and just getting all of them?

backend-api/src/functions/common/utils/environment.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type MissingEnvVarError = {
+  missingEnvVars: string[];
+};
+
+export type Config<T extends string> = {
+  [key in T]: string;
+};
+
+export const getRequiredEnvironmentVariables = <T extends string>(
```
jmooney-dd, line 16: Made another comment re envVars, would be good to chat about the general approach here. Looking at this func it feels a bit weird but unsure if it's just because of the generics and typings of stuff look a bit different to me. Not against at all but yeah, would be useful for me to chat through

backend-api/src/functions/adapters/dynamoDbAdapter.ts
chris-cooksley-gds: This is needed to appease TypeScript, annoyingly, try removing the conditions and you'll see what I mean. I can't remember the exact reason it happens, but I think essentially if you have a function that returns one (specific) type from several options, every exit point from that function needs to be returning a specific type, rather than returning an object that could be any one of the valid types.

backend-api/src/functions/adapters/dynamoDbAdapter.ts
chris-cooksley-gds: See the PR description (stray comments) for the reasoning here. tl;dr, yes, I think it's very useful for working with external calls. But not crucial to this POC, definitely something to discuss in more detail elsewhere.

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished.ts
chris-cooksley-gds: Yeah I realised that afterwards - let's use those for the real thing

backend-api/src/functions/exampleLambda/exampleLambdaConfig.ts
chris-cooksley-gds: Not sure I understand the question - this would be where we define all the env vars for the lambda, it just so happens that this example lambda only has one.

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jmooney-dd: Sorry should have re read that first, thanks for confirming!

backend-api/src/functions/common/utils/environment.ts
chris-cooksley-gds: Happy to chat through it any time (and see if it can be improved, too!). The goal was to have a generic function that could take a list of strings for the env vars we want to retrieve, and give back either a type-safe map with all these strings as keys, or an error. This way, the config functions for the individual lambdas only need to specify once which env vars they need. The pursuit of type-safety is why we've had to use all the generics etc - but there may well be a simpler way to do it.

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished.ts
sandymay-dd, line 1: General Q: What's the thinking behind some filenames being capitalised and some not? If it's intensional that is!

backend-api/src/functions/common/session/Session.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export type AuthSessionCreatedSession = CommonSessionKeys & {
+  sessionState: SessionState.AUTH_SESSION_CREATED;
+};
+
+export type BiometricTokenIssuedSession = CommonSessionKeys & {
+  sessionState: SessionState.BIOMETRIC_TOKEN_ISSUED;
+  documentType: string;
+  accessToken: string;
```
sandymay-dd, line 29: More for the back of your mind and not related to this spike - we shouldn't save the accessToken in the database - it's a short-lived token and I can't think why we'd ever need it!

backend-api/src/functions/common/session/updateOperations/ResultSent.ts
```diff
@@ -0,0 +1,17 @@
+import { IUpdateSessionOperation, SessionState } from "../Session";
+
+export class ResultSent implements IUpdateSessionOperation {
+  getDynamoDbUpdateExpression() {
+    return "set sessionState = :resultSent";
+  }
+
+  getDynamoDbConditionExpression(): undefined {
+    return undefined;
```
sandymay-dd, line 9: I assume this is undefined, given this is just a spike?

sandymay-dd: Or is this because we've already fetched the session record earlier in the lambda and we know in the lambda itself whether we want to make an update operation or not?

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished.ts
chris-cooksley-gds: The thinking was supposed to be - PascalCase for files that export one primary class or interface, with the name matching the name of said class/interface (i.e. `BiometricSessionFinished.ts` is where `BiometricSessionFinished` comes from). And camelCase for files that export a function (name matching, e.g. `exampleHandler`), or for files that export several different entities (e.g. `dynamoDBItemToSessionConvertors`). 

I don't think I've been consistent here, though - `sessionRegistry` should probably be capitalised, for example.

backend-api/src/functions/common/session/Session.ts
chris-cooksley-gds: Good point, I hadn't stopped to consider where it would actually be used again. One to update on the [TD](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3957719041/Strategic+App+ID+Check+Document+Selection+in+app) where it came from I suppose - perhaps as part of a more general review of the session schema?

backend-api/src/functions/common/session/sessionRegistry.ts
```diff
@@ -0,0 +1,30 @@
...
+  ReadSessionError,
+} from "../../adapters/dynamoDbAdapter";
+
+export interface SessionRegistry {
+  getSessionWithState(
+    sessionId: string,
+    sessionState: SessionState,
+  ): Promise<Result<Session, ReadSessionError>>;
+
+  // Ideally the return value of this would be more dynamo-agnostic, to keep it in this interface
```
sandymay-dd, line 15: Yes - I agree, and I'd prefer this. This interface would ideally return the AuthSessionCreatedSession I think?

backend-api/src/functions/common/session/updateOperations/ResultSent.ts
chris-cooksley-gds: The second one!

backend-api/src/functions/common/session/sessionRegistry.ts
chris-cooksley-gds: Yep that's what I was thinking

backend-api/src/functions/common/session/Session.ts
```diff
@@ -0,0 +1,61 @@
...
+  biometricSessionId: string;
+};
+
+export type Session =
+  | AuthSessionCreatedSession
+  | BiometricTokenIssuedSession
+  | BiometricSessionFinishedSession
+  | ResultSentSession;
+
+export interface IUpdateSessionOperation {
```
sandymay-dd, line 55: I'm just mulling (or marinadtng as Elishka would say) on this one. I like how the lambda decides what implementation of the operations to follow. 

Perhaps it's just the location of this interface! It would be good to think about partitioning (even if it's by folder structure) as much of the Dynamo implementation we can from the business logic/interfaces you've written that are so clean.

chris-cooksley-gds: Definitely happy to move the interface elsewhere. 

Something I considered was having the update operations just know about (and be able to tell you about) their valid 'FROM' session states and target session state, as well as what other fields need adding, in a dynamo-agnostic way. I just couldn't think of a neat way to work that list bit in in the time I'd taken on it. But will keep marinading.

backend-api/src/functions/common/session/Session.ts
sandymay-dd: Yes! It's on my list to update that TD having spoken to Darren

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -32,10 +55,83 @@ export class DynamoDbAdapter {
...
 
-  constructor(tableName: string) {
+  constructor(
+    tableName: string,
+    private readonly getItemToSessionConvertor = getDynamoDBItemToSessionConvertor,
+  ) {
     this.tableName = tableName;
   }
 
+  async getSessionWithState(
```
sandymay-dd, line 65: I've tried to play with a mapped type for this but probably hit similar pitfalls to what you've tried historically! If it works, it works.

PR #303, opened 2024-12-17
DCMAW-9073: Adds infrastructure and basic lambda code for /async/biometricToken

jmooney-dd:
​DCMAW-9073

### What changed
- Adds infrastructure for `/async/biometricToken` in template, is in protected subnet
- Adds lambda proxy integration to openApi spec
- Adds basic lambda code including logger dependency. Will always respond to any `POST` request with `501 Not Implemented` response 
- Adds API test for `/async/biometricToken`
- Updates infra tests to check that BiometricToken lambdas is attached to a VPC and subnets are protected

### Why did it change
Work enabling document selection in the v2 backend, see epic [here](https://govukverify.atlassian.net/browse/DCMAW-8289)

### API tests

Deployed and tested stack, all API tests (including new `501` test for `/async/biometricToken`) passing

### Manually checked logs were appearing in CloudWatch

Screenshot 2024-12-17 at 13 38 23

## Checklists


backend-api/src/functions/asyncBiometricToken/responses.ts
```diff
@@ -0,0 +1,13 @@
+import { APIGatewayProxyResult } from "aws-lambda";
+
+export const notImplementedResponse: APIGatewayProxyResult = {
```
chris-cooksley-gds, line 3: Thought: I like splitting out responses into their own file. As there's nothing specific to biometricToken about this one, might it be worth putting it in a 'common' space (e.g. in `utils/`, or in its own folder if you can think of something more specific), for ease of re-use?

backend-api/src/functions/asyncBiometricToken/responses.ts
```diff
@@ -0,0 +1,13 @@
...
+  headers: {
+    "Cache-Control": "no-store",
+    "Content-Type": "application/json",
+    "Strict-Transport-Security": "max-age=31536000",
+    "X-Content-Type-Options": " nosniff",
```
chris-cooksley-gds, line 8: Nitpick: extra space in " nosniff"

backend-api/template.yaml
```diff
@@ -982,6 +982,91 @@ Resources:
...
+      Action: lambda:InvokeFunction
+      FunctionName: !GetAtt AsyncBiometricTokenFunction.Arn
+      Principal: events.amazonaws.com
+      SourceAccount: !Sub ${AWS::AccountId}
+      SourceArn: !Sub arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${SessionsApi}/*/*/*
+
+  AsyncBiometricTokenLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      RetentionInDays: 30
```
chris-cooksley-gds, line 1030: Almost certainly beyond the scope of this ticket, as I know this is what we do elsewhere, and maybe more for @sandymay-dd  - but should we be retaining logs for 30 days, even in dev/build? That sounds expensive (although come to think if it I can't remember if they end up getting retained for 30 days anyway regardless of what we set here due to the platform setup).

backend-api/src/functions/asyncBiometricToken/responses.ts
jmooney-dd: Thanks! Removed

backend-api/template.yaml
sandymay-dd: I agree in principle. Unfortunately, it seems to be a programme requirement: https://github.com/govuk-one-login/mobile-id-check-backend/pull/1468

backend-api/template.yaml
```diff
@@ -982,6 +982,91 @@ Resources:
...
+      RoleName: !Sub ${AWS::StackName}-biometric-token-lambda
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action: sts:AssumeRole
+      Policies:
+        - PolicyName: AsyncBiometricTokenFunctionLoggingPolicy
```
sandymay-dd, line 1045: nit: should this be kebab case?

backend-api/template.yaml
chris-cooksley-gds: 😢 didn't realise it was policy, thanks for the link - seems an impractical choice for dev but I can see why it'd be hard to discriminate technically

sandymay-dd (approved):

backend-api/src/functions/asyncBiometricToken/responses.ts
chris-cooksley-gds: Happier with that! Down with dumping everything in a utils folder.

backend-api/src/functions/commonResponses/commonResponses.ts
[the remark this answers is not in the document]
sandymay-dd: Or `common/lambdaResponses` to use more of our terminology. I think I'd be more inclined to search for the word 'response' in my IDE rather than 'Result'.

backend-api/template.yaml
sandymay-dd: Only because you've made the RoleName kebab case a few lines above :)

backend-api/src/functions/commonResponses/commonResponses.ts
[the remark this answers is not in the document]
jmooney-dd: Done

backend-api/template.yaml
jmooney-dd: Agreed in standup that this approach is consistent with how this is in the rest of the template so leaving as is, tech debt ticket to be raised to fix naming conventions in future

sandymay-dd (approved):

PR #308, opened 2024-12-19
DCMAW-9046: Adds request body validation for async/biometricToken

jmooney-dd:
​DCMAW-9046

### What changed
- Adds functionality to validate request body and return required values to lambda if valid
- If request body invalid, it logs and returns `400` `invalid_request` error

### Why did it change
Ongoing work around [document selection](https://govukverify.atlassian.net/jira/software/c/projects/DCMAW/boards/612?selectedIssue=DCMAW-8289). 

### API tests
Deployed stack and ran API test (including new 400 test) and all pass

### Manually checked new log in AWS, appears as expected
Screenshot 2024-12-19 at 17 33 45

## Checklists


backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -7,17 +7,30 @@ import {
...
   dependencies: IAsyncBiometricTokenDependencies,
-  _event: APIGatewayProxyEvent,
+  event: APIGatewayProxyEvent,
   context: Context,
 ): Promise<APIGatewayProxyResult> {
   const logger = dependencies.logger();
   logger.addContext(context);
   logger.log("STARTED");
 
+  const parsedRequestBodyOrError = getParsedRequestBody(event.body);
```
jmooney-dd, line 25: Will update const and name

backend-api/src/functions/asyncBiometricToken/getParsedRequestBody/getParsedRequestBody.ts
```diff
@@ -0,0 +1,100 @@
+import { errorResult, Result, successResult } from "../../utils/result";
+
+export function getParsedRequestBody(
+  body: string | null,
+): Result<IAsyncBiometricTokenValidParsedRequestBody> {
+  if (body == null) {
+    return errorResult({
+      errorMessage: `Request body is either null or undefined. Request body: ${body}`,
+      errorCategory: "CLIENT_ERROR",
```
jmooney-dd, line 9: No need for err cat as it's validation of what client has sent, always going to be client error

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
jmooney-dd: Done

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      "Given documentType is an invalid document type",
+      {
+        sessionId: "58f4281d-d988-49ce-9586-6ef70a2be0b4",
+        documentType: "BUS_PASS",
+      },
+      "documentType in request body is invalid. documentType: BUS_PASS",
+    ],
+  ])("%s", (_description, body, errorMessage) => {
+    it("Returns an error result", () => {
+      const result = validateRequestBody(JSON.stringify(body));
```
jmooney-dd, line 69: This is why I have two describe.each FYI - I didn't want to include JSON.stringify in the params above as it would take up too much space so I do it directly in the request once. This will however fail for the 2 tests in the first describe.each block

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -7,17 +7,30 @@ import {
...
   const logger = dependencies.logger();
   logger.addContext(context);
   logger.log("STARTED");
 
+  const validateRequestBodyResult = validateRequestBody(event.body);
+  if (validateRequestBodyResult.isError) {
+    logger.log("REQUEST_BODY_INVALID", {
+      errorMessage: validateRequestBodyResult.value.errorMessage,
+    });
+    return badRequestResponse("invalid_request", "Request body invalid");
```
chris-cooksley-gds, line 30: Nitpick: since we already have more useful information about the validation failure (in `errorMessage`), I wonder if we should just send that back to the client?

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -7,17 +7,30 @@ import {
...
+  }
+  // const { sessionId, documentType } = parsedRequestBodyOrError.value
```
chris-cooksley-gds: Nitpick: comment can either be deleted or updated to reflect new variable names

backend-api/src/functions/asyncBiometricToken/registeredLogs.ts
```diff
@@ -4,8 +4,11 @@ import {
 } from "../services/logging/commonRegisteredLogs";
 import { RegisteredLogMessages } from "../services/logging/types";
 
-export type MessageName = CommonMessageNames;
+export type MessageName = "REQUEST_BODY_INVALID" | CommonMessageNames;
 
 export const registeredLogs: RegisteredLogMessages<MessageName> = {
+  REQUEST_BODY_INVALID: {
+    messageCode: "MOBILE_ASYNC_REQUEST_BODY_INVALID",
```
chris-cooksley-gds, line 11: Would `MOBILE_ASYNC_BIOMETRIC_TOKEN_REQUEST_BODY_INVALID` be better here? We'll have different endpoints validating different request bodies.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
jmooney-dd: Gone

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -1,23 +1,71 @@
...
+  beforeEach(() => {
+    mockLoggingAdapter = new MockLoggingAdapter();
+    dependencies = {
+      logger: () => new Logger(mockLoggingAdapter, registeredLogs),
+    };
+    context = buildLambdaContext();
+  });
+  describe("Request body validation", () => {
+    describe("Given request body is invalid", () => {
+      it("Logs and returns 400 Bad Request response", async () => {
```
chris-cooksley-gds, line 24: New idea (not a blocker for approval or anything): ideally, I think unit tests should aim to test just one thing at a time (and therefore to fail for just one reason), which makes catching bugs more straightforward among other things. This test checks both the logs and the lambda result. We could instead use a beforeEach or beforeAll block to invoke the function under test (and capture its result), with individual it blocks to assert on different things we care about (logs vs return value).

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -1,23 +1,71 @@
...
+  describe("Given a valid request is made", () => {
     it("Logs and returns 501 Not Implemented response", async () => {
-      const mockLoggingAdapter = new MockLoggingAdapter();
-      const dependencies = {
-        logger: () => new Logger(mockLoggingAdapter, registeredLogs),
-      };
-      const event = buildRequest();
+      const request = buildRequest({
+        body: JSON.stringify({
+          sessionId: "58f4281d-d988-49ce-9586-6ef70a2be0b4",
```
chris-cooksley-gds, line 60: Nitpick: can we extract the session ID into a variable somewhere? Even though it's just an arbitrary string at the moment, it'll come up again in the handler logic and it'll be harder to maintain a bunch of identical strings, especially when they're this complex.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
jmooney-dd: I did originally do that but then reverted as I didn't know whether it would be worth emitting info such as sessionId and documentType values here. 
- Would clients ever find use in this? 
- If not, we have that info in the logs, do we also need it here?

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.test.ts
```diff
@@ -0,0 +1,79 @@
...
+describe("Validate request body", () => {
+  describe.each([
+    [
+      "Given body is null or undefined",
+      null,
+      "Request body is either null or undefined. Request body: null",
+    ],
+    [
+      "Given body cannot be parsed as JSON",
+      "foo",
```
chris-cooksley-gds, line 12: Very pedantic nitpick/bugbear: can we use something more meaningful than 'foo' for examples? e.g. 'invalid', 'not JSON', 'arbitrary string', etc.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jmooney-dd: Done

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.test.ts
jmooney-dd: Done

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,89 @@
+import { errorResult, Result, successResult } from "../../utils/result";
+
+export function validateRequestBody(
+  body: string | null,
+): Result<IAsyncBiometricTokenValidParsedRequestBody> {
+  if (body == null) {
+    return errorResult({
+      errorMessage: `Request body is either null or undefined. Request body: ${body}`,
```
chris-cooksley-gds, line 8: Nitpick: do we need to give the value of the body/field in all of these error messages? In most cases, the message already says what it is (null, undefined, missing, empty string). It's really just the document type where it'd be particularly interesting to see the value they provided, I think.

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+  }
+
+  return successResult({
+    sessionId,
+    documentType,
+  });
+}
+
+function isAllowableDocument(
```
chris-cooksley-gds, line 70: Nitpick: `isAllowableDocumentType`?

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,89 @@
...
+  documentType: string,
+): documentType is DocumentType {
+  return (
+    documentType === "NFC_PASSPORT" ||
```
chris-cooksley-gds, line 74: Nitpick: could put the allowable types into an array and check if it includes the one provided - saves repetition and gives us one easy data structure to manage valid types.

backend-api/src/functions/asyncBiometricToken/registeredLogs.ts
jmooney-dd: Done

chris-cooksley-gds (dismissed):

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
jmooney-dd: Removed where all values are clear via message already

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
jmooney-dd: That makes sense to me, updated

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
jmooney-dd: Done

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jmooney-dd: Done

backend-api/src/functions/common/lambdaResponses.ts
```diff
@@ -1,5 +1,19 @@
 import { APIGatewayProxyResult } from "aws-lambda";
 
+export const badRequestResponse = (
+  error: string,
+  errorDescription: string,
+): APIGatewayProxyResult => {
+  return {
```
sandymay-dd, line 7: I think we need security headers here

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jmooney-dd: Have a look and see what you think now.

We will be doubling the amount of times we execute the functions this was using a beforeEach, is that a problem? 

I did try to use beforeAll, but that messed with the beforeEach in the top level block. I think the beforeAll in this block ran before the beforeEach in the top level which mean that the test failed due to things to being created in that top level beforeEach.

Feel free to improve, leave it as it is, or revert this change if you want (`c7321c704d636fa874c486923c7f0137371aa80b`)

backend-api/src/functions/common/lambdaResponses.ts
jmooney-dd: Okay I shall I add, I wasn't sure whether the security headers for the 501 were specific as listed on the ticket. Should all responses from this lambda give the same security headers?

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.test.ts
sandymay-dd, line 1: @chris-cooksley-gds -> thoughts on whether we should consider using the lambda handler to test this validation?

We've coupled these tests very slightly to our implementation -> the Result pattern. It's not bad, but made me pause for thought.

sandymay-dd: Thinking ahead, we're going to be using our Result pattern for other functions called by this lambda, so my comment doesn't have much weight.

jmooney-dd: I did this originally where this file didn't exist and I tested it in the lambda handler tests. We discussed but I think we thought it would be nicer if the lambda test file was less bloated so tested one real scenario for an invalid request body there (we didn't mock the function) and checked we give the correct status code and error etc, and then tested all invalid scenarios in this file, correct me if I'm wrong @chris-cooksley-gds

backend-api/src/functions/common/lambdaResponses.ts
jmooney-dd: Added

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jmooney-dd: Kept the testing approach just did away with the beforeEach.

backend-api/src/functions/common/lambdaResponses.ts
jmooney-dd: Added an assertion in the API test to check for them

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.test.ts
sandymay-dd: Sorry James - would have tagged you in the question, but thought you would have been on the beach by then. 🏖️ 

Agree with wanting the lambda test file to be less bloated. There are other solutions that don't change what we're testing. E.g. building test data files or having all validation tests in their own file but still testing the lambda. 

I'm 50/50 here, so happy to go with this solution. Just wanted to check whether you'd thought about the test structure, and you had 👍

sandymay-dd (approved):

PR #323, opened 2025-01-08
[DCMAW-10625] Retrieve biometric submitter access keys from SSM Parameter Store

chris-cooksley-gds:
​DCMAW-10625

### Features
* New `getSecretsFromParameterStore` adapter function, implementing a generic `GetSecrets` interface. Uses Lambda Powertools to retrieve multiple passed secrets in a single call, decrypt them, and cache them for the specified duration (configured as 15 minutes via environment variables).
* Biometric Token handler calls this adapter to retrieve submitter access keys, and selects the appropriate access key for the document type in the request
* We require these submitter keys to get an access token from ReadId. They must be stored securely in SSM Parameter store as they are secrets.

### Utility
* Introduces revamped approach to logging, currently only for Biometric Token (to be extended to other Lambdas as a near future technical improvement). This is documented at `backend-api/docs/logging.md`, but in short:
  * it uses the Powertools logger directly, without a wrapper around it. Powertools is already a powerful wrapper around the system logger, and allows us to do everything we need. Removing the wrapper brings us closer to the underlying code, and prevents the need to replicate each Powertools method we wish to use in the wrapper.
  * the logger is exported as a constant in the global scope and imported for use in any function that needs logging. This is a fairly typical usage pattern (see the [docs](https://docs.powertools.aws.dev/lambda/typescript/latest/core/logger/#resetting-keys) for example), and prevents the need to pass logger instances through chains of function calls (and updating every function signature along the way to allow this). Any persistent state is cleared at the start of each Lambda invocation.
  * a single generic logger, drawing from a single set of log messages, is used instead of lambda-specific logger instances. This ensures that shared components (like AWS adapters) can use the same logger as the rest of the Lambda function for logging. Also, grouping log messages together makes it easy to see if a pre-existing log message can be re-used, or to follow pre-existing conventions for similar messages
  * a new class (LogMessage) has been created, with locked down, pre-defined static instances for each log. This gives us type safety and easy IDE renaming.
  * The attempt and success of the network call to SSM are logged at debug level, with the lambda log level being set to debug in dev and info everywhere else. This should allow use to debug network issues without cluttering the logs in higher environments.
  * Logs are tested by spying on the console object, preventing the need for a custom mock adapter and bringing our tests closer to the actual behaviour (logging). A custom matcher has been added to facilitate this.
* Extends our usage of the Result pattern
  * Allows for custom Error objects for Failures, defaulting to the existing BaseError. **NB: to keep TypeScript happy after this change, I've had to shift to using an enum for the ErrorCategory, which is responsible for most of the files changed in this diff.**
  * Allows for empty successes and failures where no value is needed (i.e. we just need a binary succeed/fail outcome).
* Uses a slightly different pattern for retrieving config from environment variables to reduce repetition and ensure we log all missing variables. This uses a new common helper at [backend-api/src/functions/common/config/environment.ts](https://github.com/govuk-one-login/mobile-id-check-async/pull/323/files#diff-d8cc26117a109051095e58011c24092ebe747a0e96c57c42c3051232546d1204), which only needs a list of the required variables passed in.
* Uses Jest mocks for mocking dependencies in all new unit tests. Mock state is cleared for each test via the `clearMocks` option in the jest.config
* Updates eslint config to ignore unused arguments prefixed with an underscore. This is very useful when developing functions to an interface before full implementation is ready.

## Evidence

* All API tests passing against custom stack in dev:
image

* New logs:
image
image
image

## Checklists

- [X] There is a ticket raised for this PR that is present in the branch name
- [X] No PII data logged. [See guidance here](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3502407722/PII+Logging+Considerations)
- [X] Update [README](./blob/main/README.md) with any new instructions or tasks


.github/workflows/backend-api-pull-request.yml
```diff
@@ -47,7 +47,10 @@ jobs:
         run: npm run format:check
 
       # Runs unit, infra and pact tests
-      - name: Run tests 
+      - name: Run tests
+        env:
```
chris-cooksley-gds, line 51: These env vars are needed for our logging tests to succeed. The former ensures powertools will log using the console during tests, the latter ensures it will emit debug logs. I figured defining them here was simpler than running the generate-env-file script (which needs to talk to aws).

backend-api/template.yaml
```diff
@@ -155,14 +160,34 @@ Mappings:
   EnvironmentVariables:
     dev:
       STSBASEURL: 'https://mob-sts-mock.review-b-async.dev.account.gov.uk'
+      BiometricSubmitterKeySecretPathPassport: '/dev/BIOMETRIC_SUBMITTER_ACCESS_KEY_NFC_PASSPORT'
```
chris-cooksley-gds, line 163: I've defined these separately for each env (instead of just defining a static variable as 'BIOMETRIC_SUBMITTER_ACCESS_KEY_NFC_PASSPORT' and pulling in the env prefix from ${Environment}), to give us flexibility. I didn't want us to be tied to the same path format for every environment.

backend-api/tests/infra-tests/application.test.ts
```diff
@@ -460,22 +460,21 @@ describe("Backend application infrastructure", () => {
 
         expect(reservedConcurrentExecutionMapping).toStrictEqual({
           Lambda: {
-            dev: {
+            dev: expect.objectContaining({
```
chris-cooksley-gds, line 463: Had to update this as ReservedConcurrentExecutions is no longer the only key in the `Lambda` map.

.github/workflows/backend-api-pull-request.yml
sandymay-dd: Reviewing the PR later today as promised, but wanted to double check:
- is there a way to avoid the need to set environment variables here? I like how our unit test suite relies on dependency injection in the lambda constructors.
- If not, please may you double check that the containerised test suite passes? (run-tests-locally.sh)

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+  GetParametersCommand,
+  InternalServerError,
+  SSMClient,
+} from "@aws-sdk/client-ssm";
+import { emptyFailure, Result, successResult } from "../utils/result";
+import { getSecretsFromParameterStore } from "./getSecretsFromParameterStore";
+import { clearCaches } from "@aws-lambda-powertools/parameters";
+import { NOW_IN_MILLISECONDS } from "../testUtils/unitTestData";
+
+const mockSsmClient = mockClient(SSMClient);
```
chris-cooksley-gds, line 16: Testing with a mock aws sdk client (rather than jest-mocking the powertools library) gives us a looser dependency on powertools as our specific implementation for talking to SSM.

.github/workflows/backend-api-pull-request.yml
chris-cooksley-gds: POWERTOOLS_DEV will need to be set somewhere for powertools to log using the console, I think. We could possibly set it in the test files themselves but that will add an extra overhead to remember for every test file where we test logs. I'll see if I can think of any other alternatives.

chris-cooksley-gds: I've pivoted to setting them in the jest config file, so they'll apply to all tests. Seems fitting as they're both essentially part of our test configuration.

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+})
+```
+
+## Persistent Attributes
+
+Logging fields that should persist across multiple uses of the logger within a single Lambda invocation can be added as follows:
+
+```typescript
+logger.appendKeys({
+    key: value
```
sandymay-dd, line 40: nitpick: this is "key": "value" in line 29

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+})
+```
+
+As a single logger instance is shared across the modules of our Lambda, these keys will persist until cleared or
+overwritten. Request-specific keys should be cleared at the start of each request by calling `logger.resetKeys()`
```
sandymay-dd, line 45: suggestion[take it or leave it]:
```suggestion
overwritten. Request-specific keys should be cleared at the start of each Lambda invocation by calling `logger.resetKeys()` 
```
Avoids having request twice in the sentence in different usages (compound adjective and a noun).

backend-api/src/functions/common/logging/logger.ts
```diff
@@ -0,0 +1,9 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { Context } from "aws-lambda";
+
+export const logger = new Logger();
+
+export function setupLoggerForNewInvocation(logger: Logger, context: Context) {
+  logger.resetKeys();
+  logger.addContext(context);
+}
```
sandymay-dd, line 9: suggestion[take it or leave it]: Remove the utility, despite its usefulness.  I'm not a huge fan of a mutation function/one that updates the state of the argument passed in. This is a very mild feeling, if you're happy here, leave it.

backend-api/src/functions/jwks/configService/configService.ts
```diff
@@ -12,17 +17,17 @@ export class ConfigService implements IGetConfig<Config> {
     if (!env.ENCRYPTION_KEY_ID)
       return errorResult({
         errorMessage: "No ENCRYPTION_KEY_ID",
-        errorCategory: "SERVER_ERROR",
+        errorCategory: ErrorCategory.SERVER_ERROR,
```
sandymay-dd, line 20: A good change

backend-api/src/functions/jwks/jwksBuilder/tests/mocks.ts
```diff
@@ -32,7 +37,7 @@ export class MockJwksBuilderErrorResult implements IJwksBuilder {
     return Promise.resolve(
       errorResult({
         errorMessage: "Error formatting public key as JWK",
-        errorCategory: "SERVER_ERROR",
+        errorCategory: ErrorCategory.SERVER_ERROR,
```
sandymay-dd, line 40: For marination [no change suggested]: There's starting (as I scroll through the PR) to be some duplication here on returning an error result with a server error + error message

backend-api/src/functions/services/session/tests/sessionService.test.ts
```diff
@@ -38,7 +39,7 @@ describe("Session Service", () => {
 
         expect(result.value).toStrictEqual({
           errorMessage: "Error getting session ID - Error: Mock DB Error",
-          errorCategory: "SERVER_ERROR",
+          errorCategory: ErrorCategory.SERVER_ERROR,
```
sandymay-dd, line 42: passing thought [no action suggested]: If there weren't so many instances of the errorCategory in our codebase, I'd have suggested keeping this hardcoded. I don't think any tests would fail if someone updated the ErrorCategory.SERVER_ERROR string.

sandymay-dd (commented):
Review has been interrupted. Reviewed Readme and the refactor of the errorCategory thus far.

backend-api/src/functions/services/session/tests/sessionService.test.ts
sandymay-dd: Could be wrong - I haven't read through the whole PR.

backend-api/docs/logging.md
chris-cooksley-gds: Good catch, have standardised

backend-api/docs/logging.md
chris-cooksley-gds: Agreed and done

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+or calling the `setupLoggerForNewInvocation()` utility, which will invoke the former.
+
+## Log Levels
+
+The log level for all of our Lambda functions is configured using the `Globals.Function.LoggingConfig` in our
+`template.yaml`. Currently, the log level is set to DEBUG in our dev environment, and INFO in every other environment.
+This means that only logs with a level of DEBUG or above will be visible in dev, and only INFO or above in all other
+environments.
+
+Most of our informational logs - such as logs for the attempt and success of each network call - are logged at debug
```
sandymay-dd, line 54:
```suggestion
Most of our informational logs - such as logs for the attempt and success of each network call - are logged at DEBUG 
```

backend-api/src/functions/jwks/jwksBuilder/tests/mocks.ts
chris-cooksley-gds: I think at some point we should go through and:
* identify any places where we don't need/use the error message/category (there may not be any, but I wouldn't be surprised if in some scenarios we just care whether it was succeeded and failed), and replace with an empty failure
* introduce helpers if needed for quickly generating failures with server/client BaseErrors.

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+Typically, the only non-error logs emitted at INFO level are the 'STARTED' and 'COMPLETED' logs for each Lambda function,
+which allow us to easily track successful vs unsuccessful function completions.
+
+For more information on configuring Lambda logs, see the [official documentation](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-logging.html).
+
+## Personally Identifiable Information
+
+Personally identifiable information (PII) is any information that can be used to identify an individual, either alone 
+or when combined with other information. We should be careful to **never log PII**, even at debug level, and to never
+allow code that could log PII into version control.
```
sandymay-dd, line 69: suggestion: Please may we add something to say that we should be careful when logging out entire objects and should avoid this where possible. It is possible that the object changes over time and may include PII which would then be included in an existing log

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+```
+fields @timestamp, message, messageCode
+| filter messageCode = 'MOBILE_ASYNC_GET_SECRETS_FROM_PARAMETER_STORE_FAILURE'
+| sort @timestamp desc
+| limit 20
+```
+
+## Testing Logs
+
+Unit tests for log messages can be achieved using our custom matcher (defined in `backend-api/src/functions/testUtils/matchers.ts`).
```
sandymay-dd, line 90: nitpick
```suggestion
There is a custom matcher (defined in `backend-api/src/functions/testUtils/matchers.ts`) that can be used to facilitate unit testing log messages.
```
or something similar

backend-api/docs/logging.md
sandymay-dd, line 1: suggestion[take it or leave it]: Is it possible to add a section somewhere that indicates all lambdas must have a STARTED and COMPLETED log to help in monitoring the success of a lambda etc.?

backend-api/src/functions/services/session/tests/sessionService.test.ts
chris-cooksley-gds: I was on the fence about this - normally (e.g. for things like log message codes) I don't like to refer to entities from the application code in the test assertions, for the reason you mention, we'd be unprotected from mutations of the underlying string if both application and tests get the value from the same place. But in this case as you point out we're referencing it in so many different places, that using a hardcoded string would make it quite tedious to change, so I figured referencing the enum was the lesser evil.

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+        Parameters: [
+          {
+            Name: mockSecretName1,
+            Value: mockSecretValue1,
+          },
+        ],
+        InvalidParameters: [mockSecretName2],
+      });
+      result = await getSecretsFromParameterStore({
+        secretNames: [mockSecretName1],
```
sandymay-dd, line 91: [question]: I'm getting a bit confused here. The test is requesting `mockSecretName1` but the mock has been configured to also return `mockSecretName2` in the InvalidParameters section.

If I've looked at the AWS docs correctly, would it be better to add `mockSecretName2` to the function under test, i.e.:

`result = await getSecretsFromParameterStore({
        secretNames: [mockSecretName1, mockSecretName2],
});`

?

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+
+      it("Logs success at debug level", () => {
+        expect(consoleDebugSpy).toHaveBeenCalledWithLogFields({
+          messageCode: "MOBILE_ASYNC_GET_SECRETS_FROM_PARAMETER_STORE_SUCCESS",
+        });
+      });
+
+      it("Returns success with values of secrets in same order", () => {
+        expect(result).toEqual(
+          successResult([mockSecretValue1, mockSecretValue2]),
```
sandymay-dd, line 190: question: For the caller of this adapter, there's some implicit knowledge that secrets are returned in the same order in which they were sent. Should we think about returning key:value pairs here?

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+        );
+      });
+    });
+  });
+
+  describe("When called twice for same path within cache duration", () => {
```
sandymay-dd, line 196: question: What does path mean here please? It's not a term we use in the rest of the tests

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+        cacheDurationInSeconds: 10,
+      });
+
+      jest.setSystemTime(NOW_IN_MILLISECONDS + 10000);
+
+      result = await getSecretsFromParameterStore({
+        secretNames: [mockSecretName1, mockSecretName2],
+        cacheDurationInSeconds: 10,
+      });
+    });
```
sandymay-dd, line 227: question[I'm possibly confused]: We're advancing time by 10000 milliseconds as soon as we've made the first SSM SDK request. The cache duration is 10 seconds. Does this not exceed the cache duration?

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
sandymay-dd, line 1: Excellent readability of tests. It was a pleasure to look through, if one draws pleasure from reading unit tests.

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
chris-cooksley-gds: That's what I'd expected, but no - powertools considered the cache valid up to and including 10 seconds from the previous request, and invalid a millisecond after.

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+## Log Levels
+
+The log level for all of our Lambda functions is configured using the `Globals.Function.LoggingConfig` in our
+`template.yaml`. Currently, the log level is set to DEBUG in our dev environment, and INFO in every other environment.
+This means that only logs with a level of DEBUG or above will be visible in dev, and only INFO or above in all other
+environments.
+
+Most of our informational logs - such as logs for the attempt and success of each network call - are logged at debug 
+level. This will make it easier to debug issues where they can be reproduced in dev (which is most of the time), without
+emitting too many redundant logs to CloudWatch in higher environments. If more information is needed to debug a
```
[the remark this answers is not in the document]
chris-cooksley-gds: Changed to non-error to make it marginally less janky.

backend-api/docs/logging.md
chris-cooksley-gds: Done

backend-api/docs/logging.md
chris-cooksley-gds: Done

backend-api/docs/logging.md
chris-cooksley-gds: Done

backend-api/docs/logging.md
chris-cooksley-gds: It's mentioned indirectly in the log level section, but I've added an explicit section on standard log messages now.

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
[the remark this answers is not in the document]
chris-cooksley-gds: I've seen it done both ways across the codebase (test at same level vs test in tests/ subdirectory), but happy to align on the subdirectory approach if that's preferred. I'm still adjusting to having the unit tests next to the source code in the first place (which I have come to enjoy)!

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
chris-cooksley-gds: You're right, I must have forgot to send the second secret name. Added now.

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
```diff
@@ -0,0 +1,294 @@
...
+      expect(result).toEqual(
+        successResult([mockSecretValue1, mockSecretValue2]),
+      );
+    });
+
+    it("Makes only one call to SSM", () => {
+      expect(mockSsmClient).toHaveReceivedCommandTimes(GetParametersCommand, 1);
+    });
+  });
+  describe("When called twice for same path beyond cache duration", () => {
```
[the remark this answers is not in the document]
chris-cooksley-gds: As above

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
chris-cooksley-gds: I'd thought about doing this and left it as a list in case I was overthinking it. I'll look into it again.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -1,27 +1,119 @@
...
+  let result: APIGatewayProxyResult;
+
+  const validRequest = buildRequest({
+    body: JSON.stringify({
+      sessionId: mockSessionId,
+      documentType: "NFC_PASSPORT",
+    }),
+  });
+
+  const mockSuccessfulGetSecrets = jest
```
sandymay-dd, line 30: nit[take it or leave it]:
```suggestion
  const mockGetSecretsSuccess = jest
```

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -33,25 +125,19 @@ describe("Async Biometric Token", () => {
         }),
       });
 
-      it("Logs BIOMETRIC_TOKEN_REQUEST_BODY_INVALID", async () => {
-        await lambdaHandlerConstructor(dependencies, request, context);
+      beforeEach(async () => {
```
sandymay-dd, line 128: question[I'm curious]: Is there a reason we've not moved the request setup into the beforeEach?

```
const request = buildRequest({
        body: JSON.stringify({
          sessionId: mockSessionId,
          documentType: "BUS_PASS",
        }),
      });
```

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -4,35 +4,94 @@ import {
...
+    logger.error(LogMessage.BIOMETRIC_TOKEN_REQUEST_BODY_INVALID, {
       errorMessage,
     });
     return badRequestResponse("invalid_request", errorMessage);
   }
+  const documentType = validateRequestBodyResult.value.documentType;
+
+  const submitterKeyResult = await getSubmitterKeyForDocumentType(
+    documentType,
+    config,
```
sandymay-dd, line 53: suggestion[take it or leave it]: Explicitly declare what is being passed into the function call for readability

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -4,35 +4,94 @@ import {
...
+      config.BIOMETRIC_SUBMITTER_KEY_SECRET_PATH_DL,
+    ],
+    cacheDurationInSeconds: Number(
+      config.BIOMETRIC_SUBMITTER_KEY_SECRET_CACHE_DURATION_IN_SECONDS,
+    ),
+  });
+  if (getSubmitterKeysResult.isError) {
+    return emptyFailure();
+  }
+  const [submitterKeyPassport, submitterKeyBrp, submitterKeyDl] =
```
sandymay-dd, line 87: suggestion: I'd rather keep the knowledge that Powertools/SSM returns back secrets in the order they were requested to the adapter. Slightly nervous that there's an accidental change in order on this line and we give the mobile apps the wrong access token (there was an incident on this a long time)

backend-api/src/functions/asyncBiometricToken/biometricTokenConfig.ts
sandymay-dd, line 1: thank you for introducing this.

backend-api/template.yaml
sandymay-dd: question: `I didn't want us to be tied to the same path format for every environment.` -> what issue would this cause?

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
chris-cooksley-gds: Could you elaborate on this one? Do you mean splitting out the specific items of config we're passing in rather than passing the full config object?

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -65,32 +151,55 @@ describe("Async Biometric Token", () => {
     });
   });
 
-  describe("Given a valid request is made", () => {
-    const request = buildRequest({
-      body: JSON.stringify({
-        sessionId: mockSessionId,
-        documentType: "NFC_PASSPORT",
-      }),
+  describe("When there is an error getting secrets", () => {
```
sandymay-dd, line 152: nitpick[take it or leave it]: Very subtly different naming style to the two previous describe blocks.

image

backend-api/src/functions/adapters/getSecretsFromParameterStore.test.ts
chris-cooksley-gds: I've refactored to have it return key value pairs. I couldn't think of a nice way to make typescript guarantee that all/only the secret names passed in would exist as keys in the returned object - but I've added a check to make sure that all of the secrets requested are actually present in the returned object. Let me know what you think.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
chris-cooksley-gds: Done

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
chris-cooksley-gds: Nope, must have missed it when updating the existing tests. Done.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
chris-cooksley-gds: Think I'll leave this one as is, on the basis that the first two describe blocks are just groupings of related 'Given XYZ' tests to help with organisation where there are several related tests. But the latter two describe blocks don't need to be sub-grouped in this way (there's only one test for a secrets error and one for a happy path), so they sit at the top level. All of the lowest-level describe blocks (immediately surrounding an it block) follow the same convention though, except for a bit of inconsistency on given vs when.

(Obvs always happy to discuss further)

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
sandymay-dd: Yep, exactly that. From reading the PR, I wasn't sure exactly what config was being used in the `getSubmitterKeyForDocumentType` function. 

```
const submitterKeyResult = await getSubmitterKeyForDocumentType(
    documentType,
    getBiometricSubmitterSecretKeyPaths(config),
    dependencies.getSecrets,
  );
```
🤷 

Again - take or leave the comment. It's absolutely fine as-is.

.github/workflows/backend-api-pull-request.yml
sandymay-dd: Thank you. Happy with this.

backend-api/src/functions/testUtils/matchers.ts
```diff
@@ -0,0 +1,47 @@
...
+  return Object.keys(object).every((key) =>
+    deepEquals(object[key], targetObject[key]),
+  );
+}
+
+function deepEquals(subject: unknown, target: unknown): boolean {
+  return JSON.stringify(subject) === JSON.stringify(target);
+}
+
+expect.extend({
```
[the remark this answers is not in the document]
chris-cooksley-gds: Done

backend-api/docs/logging.md
```diff
@@ -0,0 +1,122 @@
...
+
+```typescript
+logger.appendKeys({
+    key: value
+})
+```
+
+As a single logger instance is shared across the modules of our Lambda, these keys will persist until cleared or
+overwritten. Request-specific keys should be cleared at the start of each request by calling `logger.resetKeys()` 
+or calling the `setupLoggerForNewInvocation()` utility, which will invoke the former.
```
sandymay-dd: issue[I think!]: I think this utility has now been removed so this sentence is out of date.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
chris-cooksley-gds: I think on balance I'll leave this one as is, as we need four different bits of config in the helper (three paths and a cache duration), and I'd rather not add the complexity to the interface if we already have a typesafe object with all the keys on it.

sandymay-dd (approved):
💯

backend-api/docs/logging.md
chris-cooksley-gds: I think this has already been removed in the latest version of the file (I took it out when I inlined the function).

backend-api/template.yaml
chris-cooksley-gds: Just inflexibility. We happen to be planning to use the same path structure in each environment, but if we ever wanted to do things differently for an environment (e.g. if there was a specific naming convention we needed to follow in prod), this would make it easy rather than hard. Perhaps not very likely in this situation, but I suppose it's more of a general guideline that I try to follow for environment configuration.

sandymay-dd: Makes sense and thanks for the explanation. I've just scanned the template.yaml to check whether this introduces an 'inconsistency' - it would be nice to update the client registry secret name in the template in the same vain? 

`${Environment}/clientRegistry`

I'll raise a tiny ticket so we remember to do it Monday if you're happy

PR #324, opened 2025-01-09
[DCMAW-10897] Async CloudFront | Update helper scripts to deploy cloudfront distribution in dev

dothomson:
To prepare for the cloudfront distribution in async this change adds the deployment to the helper scripts. 

Since the stack is taken from devplatform then it uses the `aws cloudformation` command instead of the `sam -` commands.

The template version is currently locked to v1.8.0 which may drift from the version deployed via https://github.com/govuk-one-login/mobile-id-check-async-infra 

In generating the parameters for the templates steps are taken to handle sensitive values with care. The `OriginCloakingHeader` is randomly generated if the stack currently doesn't exist, otherwise the current value is retrieved. The ACM certificate UUID is pulled down each time.

Test:

The script has successfully deployed all stacks from scratch: 
 
image

If the script is run a second time with no changes then it also succeeds.

### DIFF
--- a/helper-scripts/.gitignore
+++ b/helper-scripts/.gitignore
@@ -1,1 +1,0 @@
-parameters-*-cf-dist.json
--- a/helper-scripts/.gitignore
+++ b/helper-scripts/.gitignore
@@ -1,0 +1,1 @@
+parameters-*-cf-dist.json
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,9 @@ docker-vars.env
 
 async-proxy-private-spec.yaml
 
+# parameters files generated from the helper-script deployments of the cloudfront distribution
+parameters-*-cf-dist.json
+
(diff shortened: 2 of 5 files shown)


helper-scripts/.gitignore
```diff
@@ -0,0 +1 @@
+parameters-*-cf-dist.json
```
kikidawson-gds, line 1: Can we have one .gitignore at the root? Only because I find it neater but open to discussion

chris-cooksley-gds (dismissed):

helper-scripts/.gitignore
dothomson: I have followed your suggestion.

chris-cooksley-gds (approved):

helper-scripts/.gitignore
kikidawson-gds: You have a point though

PR #325, opened 2025-01-09
[DCMAW-10899] Async CloudFront | Route backend traffic through the cloudfront distribution

dothomson:
Reroutes the traffic for the Session API through the cloudfront distribution stack. The CF DIstribution template is provided by Dev Platform and is now a requirement for the backend api stack in all instances. 

Changes: 

- `${VANITY_URL}` (i.e. `sessions-mob-async-backend.review-b-backend.dev.account.gov.uk`) now points to the cloudfront distribution domain
- Creates the `origin.${VANITY_URL}` route53 record which targets the sessions api gateway domain name. 
- Attaches the origin protection waf to ensure all traffic is passing through the cloudfront distribution. 
- Preemptively adds the tag `FMSRegionalPolicy: false` to exclude the api gateway stage from the automated WAF association and preserve the origin protection WAF.

Test: 

- Deployed a backend stack `douglast-async-backend` which passes test successfully. 
- Deployed a `douglast-async-backend-cf-dist` stack. 

- Created a change set for `douglast-async-backend` with the template changes in this PR:
image

- Executed the change set 
- Reran tests `sh ./run-tests-locally.sh douglast-async-backend`
image
image

- Checked metrics and CloudFront is seeing the same amount of requests as the API Gateway: 
image

- curl commands: I used the following script to prove that access is blocked through the `origin.${VANITY_URL}` without the origin cloaking header:

```
vanity_url=sessions-douglast-async-backend.review-b-async.dev.account.gov.uk
origin_url="origin.${vanity_url}"

vanity_ip=$(dig +short "${vanity_url}" | head --lines 1)
origin_ip=$(dig +short "${origin_url}" | head --lines 1)

echo "Confirm origin ip and vanity ip are different:"
echo "Vanity IP: ${vanity_ip}"
echo "Origin IP: ${origin_ip}"
echo

echo "Confirm connection successfull to vanity and origin WITHOUT the origin-cloaking-secret header"
echo "VANITY:"
curl --verbose --resolve "${vanity_url}:443:${vanity_ip}" "https://${vanity_url}/.well-known/jwks.json" --output response_vanity.txt --silent

echo
echo

echo "ORIGIN:"
curl --verbose --resolve "${vanity_url}:443:${origin_ip}" "https://${vanity_url}/.well-known/jwks.json" --output response_origin.txt --silent

echo
echo

echo "Response body diff:"
echo "------------"
diff response_vanity.txt response_origin.txt
echo "------------"
```

### DIFF
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -811,7 +811,7 @@ Resources:
           Parameters:
             Location: ./openApiSpecs/async-public-spec.yaml
       Tags:
-        FMSRegionalPolicy: false
+        FMSRegionalPolicy: false # Disables FMS for this API Gateway in favor of the cf-dist WAF. FMS Docs: https://govukverify.atlassian.net/wiki/x/eoBNEQE
 
   SessionsApiAccessLogs:
     Type: AWS::Logs::LogGroup
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -875,9 +875,6 @@ Resources:
         DNSName:
           Fn::ImportValue: !Sub "${AWS::StackName}-cf-dist-DistributionDomain"
         HostedZoneId: Z2FDTNDATAQYW2
-        # AliasTarget:
-        #   DNSName: !GetAtt SessionsApiDomainName.RegionalDomainName
-        #   HostedZoneId: !GetAtt SessionsApiDomainName.RegionalHostedZoneId
 
   SessionsApiWebAclAssociation:
     Type: AWS::WAFv2::WebACLAssociation
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -810,6 +810,8 @@ Resources:
           Name: AWS::Include
           Parameters:
             Location: ./openApiSpecs/async-public-spec.yaml
+      Tags:
+        FMSRegionalPolicy: false
 
   SessionsApiAccessLogs:
     Type: AWS::Logs::LogGroup
--- a/backend-api/template.yaml
+++ b/backend-api/template.yaml
@@ -212,7 +212,7 @@ Conditions:
       - !Equals
         - !Ref Environment
         - dev
-    - !Equals [ !Ref DeployAlarmsInDev, true ]
+    - !Equals [!Ref DeployAlarmsInDev, true]
 
 Globals:
   Function:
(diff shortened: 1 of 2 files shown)


mthkbarber (dismissed):

backend-api/template.yaml
```diff
@@ -1430,8 +1457,7 @@ Resources:
       ActionsEnabled: true
       AlarmDescription: !Sub
         - "1% or more of requests are failing on the /.well-known/jwks.json endpoint. See runbook: ${RunbookUrl}"
-        - RunbookUrl:
-            !FindInMap [StaticVariables, urls, WarningAlarmsRunbook]
+        - RunbookUrl: !FindInMap [StaticVariables, urls, WarningAlarmsRunbook]
```
sandymay-dd, line 1460: I smell rain

backend-api/template.yaml
```diff
@@ -857,8 +874,18 @@ Resources:
       Type: A
       HostedZoneId: !Sub '{{resolve:ssm:/${Environment}/Platform/Route53/PrimaryZoneID}}'
       AliasTarget:
-        DNSName: !GetAtt SessionsApiDomainName.RegionalDomainName
-        HostedZoneId: !GetAtt SessionsApiDomainName.RegionalHostedZoneId
+        DNSName:
+          Fn::ImportValue: !Sub "${AWS::StackName}-cf-dist-DistributionDomain"
+        HostedZoneId: Z2FDTNDATAQYW2
```
sandymay-dd, line 879: question: If this is hardcoded, won't it differ across environments?

backend-api/template.yaml
```diff
@@ -810,6 +810,8 @@ Resources:
           Name: AWS::Include
           Parameters:
             Location: ./openApiSpecs/async-public-spec.yaml
+      Tags:
+        FMSRegionalPolicy: false
```
sandymay-dd, line 814: suggestion: I'd be quite keen on adding an infra test for this to help document that this tag with a comment in the test that conveys:

isn't just for adding metadata to a template. It is used by FMS to determine whether to associate a central Web ACL. This shouldn't be removed etc.

sandymay-dd: Slightly nervous that someone might remove it one day in an attempt to standardise tags. Unlikely I know, but I'd rather be overcautious

backend-api/template.yaml
mthkbarber: It's a fixed value according to the docs https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset-aliastarget.html#:~:text=Specify%20Z2FDTNDATAQYW2.%20This%20is

backend-api/template.yaml
```diff
@@ -845,6 +847,21 @@ Resources:
...
+            - DNS
+            - !Ref Environment
+            - BaseDns
+      Type: A
+      HostedZoneId: !Sub "{{resolve:ssm:/${Environment}/Platform/Route53/PrimaryZoneID}}"
+      AliasTarget:
+        DNSName: !GetAtt SessionsApiDomainName.RegionalDomainName
+        HostedZoneId: !GetAtt SessionsApiDomainName.RegionalHostedZoneId
+
   SessionsApiRecordSet:
```
mthkbarber, line 865: Is it wrong of me to think that it would be helpful if it referenced CloudFront or not?

sandymay-dd (dismissed):

sandymay-dd (dismissed):

backend-api/template.yaml
dothomson: A test and a comment have been added.

backend-api/template.yaml
dothomson: Its a good instinct but I can't figure out a clean name to change it to. I think the reference in `AliasTarget` makes it clear what this resource is doing.

sandymay-dd (approved):

PR #328, opened 2025-01-13
DCMAW-10844 : Organise async backend resources into sub-templates

tmulberry-gds:
​DCMAW-10844

### What changed
Organised async backend resources into sub-templates. Created script to merge into single template.yaml file

### Why did it change
original template.yaml file was getting too large and difficult to keep track of and update. Splitting into manageable chuncks allows for easier maintenance and modularity.

## Checklists

image
image


.pre-commit.config.yaml
chris-cooksley-gds, line 1: Is this superseded by the updates to the pre-commit file above? It looks like it would run the same script (albeit looking for it in a different directory), and presumably requires the [pre-commit](https://pre-commit.com/) framework

.github/hook-scripts/pre-commit
```diff
@@ -7,8 +7,20 @@
...
+echo "Running lint and format checks..."
 npm run lint
-npm run format:check
\ No newline at end of file
+npm run format:check
+
+echo "Verifying CloudFormation templates..."
+# Change back to root directory to run template verification
+cd "$GIT_ROOT" || exit 1
+"$GIT_ROOT/helper-scripts/merge-templates.script.sh" verify
```
chris-cooksley-gds, line 23: Question for my own interest: what does this construction ("<command>" verify) do?

backend-api/infra/storage/dynamodb.cloudFormation.yaml
chris-cooksley-gds, line 1: Nitpick: can the name for this relate to the domain/purpose of the resources within (i.e. something like `session` or `sessions-table` etc), rather than the particular AWS resource used? I think it's more likely that we'll want to access, say, all the infrastructure related to storing sessions, rather than all of the dynamodb infrastructure.

Ditto for our s3 and sqs stuff.

backend-api/infra/monitoring/alarms.cloudFormation.yaml
```diff
@@ -0,0 +1,76 @@
...
+                - Name: Resource
+                  Value: /.well-known/jwks.json
+            Period: 60
+            Stat: Sum
+        - Id: errorPercentage5XX
+          Label: "Number of 5XX errors returned as a percentage"
+          ReturnData: false
+          Expression: (errorSum5XX/invocations) * 100
+
+  # Additional alarms would be defined here following the same pattern
```
chris-cooksley-gds, line 75: Is there a reason we've not added the other alarms here yet?

Separately - I think there's value in not keeping all of our alarms in the one template, as it will become very large very quickly. What would you think about splitting them by endpoint?

backend-api/template.old.yaml
chris-cooksley-gds, line 1: Is this to be deleted?

.github/hook-scripts/pre-commit
chris-cooksley-gds: Nevermind, I've read the script now, it's an argument. Followup question, then - would it be more useful to have the pre-commit hook go ahead and run the merge, rather than just prompting the user to do it themselves? Can't think of any situations where we're committing and wouldn't want to run the merge. (I realise we only prompt the user for linting/formatting at the moment so not a dealbreaker)

backend-api/infra/api/private-api.cloudFormation.yaml
chris-cooksley-gds, line 1: Pedantic nitpick: I like clarifying the type of yaml file in the filename like this. But can we use consistent casing with that and the main file name (i.e. privateApi.cloudFormation, private-api.cloud-formation, private-api.cloudformation)

backend-api/infra/functions/active-session.cloudFormation.yaml
chris-cooksley-gds, line 1: Question for discussion (not an issue): I can see that you've differed from the STS implementation linked on the ticket, in that the sub-templates are brought in as stacks with all config passed in as parameters, and we use a rain fmt instead of a rain merge to bring them together.

I suppose the biggest benefit of this approach is linting/reference safety, in that every bit of config we reference is explicitly passed in as a parameter (whereas in the STS implementation we have to just reference what's defined in the parent and deal with the red underlines, and wouldn't catch a typo). 

But on the other hand, it makes it more cumbersome to define new sub-templates or bits of config, as we will always need to update the parent to pull in a new stack and define any parameters, and define new parameters in the sub-templates.

What swung it for you? Are there any other factors I hadn't considered?

helper-scripts/merge-templates.script.sh
chris-cooksley-gds, line 1: Something for us all to remember for later (I don't think it necessarily needs addressing now) - this script is currently set to work only for the backend-api (as per the ticket). But I imagine we'll eventually want to use the same pattern for our test-resources/sts-mock stacks, or any other stacks we might introduce to this monorepo in future. So we'll probably want a way to run this easily in different stack directories.

.github/CODEOWNERS
```diff
@@ -9,3 +9,6 @@
 # All changes to Dockerfiles and CloudFormation templates to be checked by SRE
 **/template.yaml @govuk-one-login/mobile-id-check-sre @govuk-one-login/mobile-leads
 **/Dockerfile @govuk-one-login/mobile-id-check-sre @govuk-one-login/mobile-leads
+
+# Platform team owns CloudFormation templates
+/infra/ @govuk-one-login/mobile-leads
```
dothomson, line 14:
```suggestion
infra/ @govuk-one-login/mobile-id-check-sre @govuk-one-login/mobile-leads
```

backend-api/infra/api/private-api.cloudFormation.yaml
kikidawson-gds: Suggesting `private.cloudFormation` as we're already in a directory called api. Cloudformation is one word so I think we can keep that as it is.

.github/hook-scripts/pre-commit
tmulberry-gds: Resolved

backend-api/infra/api/private-api.cloudFormation.yaml
tmulberry-gds: Resolved

backend-api/infra/functions/active-session.cloudFormation.yaml
sandymay-dd: @tmulberry-gds FYI in case this one got lost

tmulberry-gds:
Closing this PR as changing the way of how this will be implemented

PR #336, opened 2025-01-14
[DCMAW-11018] Use new certificates

kikidawson-gds:
[DCMAW-11018](https://govukverify.atlassian.net/browse/DCMAW-11018)

### What changed
Updates regional certs to point to either wildcard cert if dev, or proxy/base cert

### Why did it change
Current set up is fiddly to add new domains.

## Checklists


backend-api/template.yaml
```diff
@@ -224,6 +224,10 @@ Conditions:
       - !Ref Environment
       - build
 
+  useWildcardCertificate: !Equals
```
dothomson, line 227: Could also check that `${AWS::StackName} != 'mob-async-backend'` so that the auto dev deployment uses the BaseCertificate.

kikidawson-gds: Does this mean that if the stack name isn't mob-async-backend then it'll add a wildcard. Is that a bit open?

dothomson: Only if it is also dev! 
```
!And
  - !Equals [!Ref Environment, "dev]
  - !Not [!Equals [!Ref AWS::StackName, "mob-async-backend"]]
```

or something

kikidawson-gds: OH makes sense. Yes, love this. Will update

backend-api/template.yaml
```diff
@@ -659,7 +663,11 @@ Resources:
       EndpointConfiguration:
         Types:
           - REGIONAL
-      RegionalCertificateArn: !Sub '{{resolve:ssm:/${Environment}/Platform/ACM/AsyncPrimaryZoneWildcardCertificateARN}}'
+      RegionalCertificateArn: !If
+        - UseWildcardCertificate
+        - !Sub '{{resolve:ssm:/${Environment}/Platform/ACM/AsyncPrimaryZoneWildcardCertificateV1ARN}}'
```
sandymay-dd, line 668: What does the V1 signify here @kikidawson-gds?

sandymay-dd (approved):

backend-api/template.yaml
sandymay-dd: Approved assuming it's some missing knowledge on my part and it's not a reference to the V1 vs V2 AWS account set

kikidawson-gds: I added some notes about this in the infra readme. Do you think I need to pad it out a little more? Maybe I should put an example
https://github.com/govuk-one-login/mobile-id-check-async-infra/pull/13/files

sandymay-dd: I literally just saw that on the next PR review :) happy with how it is now

PR #337, opened 2025-01-15
DCMAW-10829: Get biometric access token from ReadID

jmooney-dd:
### What changed
- Adds functionality to `getBiometricToken()` to send request to ReadID's `/token` endpoint with `submitterKey` to retrieve biometric access token
- Adds `POST` functionality in `sendHttpRequest()`
- Updates `template.yaml` to use v2 of ReadID mock
- Adds logs when making request to ReadID for biometric access token:
  - `MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT`
    - `level: "debug"` - will only appear in `dev`
  - `MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE`
    -  `level: "error"` - will appear in all environments
  - `MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_SUCCESS`
    - `level: "debug"` - will only appear in `dev`

### Why did it change
ReadID host a `/token` endpoint. This used for exchanging the submitter access key for a temporary access key that is scoped for use by the mobile apps. The submitter access keys have been retrieved 

### Next steps

This PR covers the logic required to get a biometric access token from ReadID however a follow-up PR will be raised to adapt `sendHttpRequest()` to allow for an exponential backoff retry policy, currently it is static

### Manual testing

First time integrating with our ReadID Mock in this account so made a real request to it and confirmed I received the expected response from the mock

### Automation testing

Deployed stack and ran `api` tests - all passing

### New log messages

`MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT`

Screenshot 2025-01-15 at 23 13 44

`MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE`

Screenshot 2025-01-16 at 00 04 55

`MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_SUCCESS`

Screenshot 2025-01-15 at 23 14 16

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
@@ -40,11 +40,11 @@ describe("getBiometricToken", () => {
     });
   });
 
-  describe("Given there is an error making network request", () => {
+  describe("Given there is an error when making network request", () => {
     beforeEach(async () => {
       mockSendHttpRequest = jest
         .fn()
-        .mockRejectedValue(new Error("Unexpected network error"));
+        .mockRejectedValue(new Error("mockError"));
 
       result = await getBiometricToken(
         "https://mockUrl.com",
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
@@ -14,11 +14,12 @@ describe("getBiometricToken", () => {
 
   describe("On every call", () => {
     beforeEach(async () => {
-      global.fetch = jest.fn(() =>
-        Promise.resolve(),
-      ) as jest.Mock;
+      global.fetch = jest.fn(() => Promise.resolve()) as jest.Mock;
 
-      result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
+      result = await getBiometricToken(
+        "https://mockUrl.com",
+        "mockSubmitterKey",
+      );
     });
 
     it("Logs network call attempt at debug level", () => {
@@ -26,8 +27,8 @@ describe("getBiometricToken", () => {
         messageCode:
           "MOBILE_ASYNC_BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT",
       });
-    })
-  })
+    });
+  });
 
   describe("Given there is an error making network request", () => {
     beforeEach(async () => {
@@ -35,7 +36,10 @@ describe("getBiometricToken", () => {
         Promise.reject(new Error("Unexpected network error")),
       ) as jest.Mock;
 
-      result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
+      result = await getBiometricToken(
+        "https://mockUrl.com",
+        "mockSubmitterKey",
+      );
     });
 
     it("Logs error", () => {
@@ -68,7 +72,10 @@ describe("getBiometricToken", () => {
           } as unknown as Response),
         ) as jest.Mock;
 
-        result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
+        result = await getBiometricToken(
+          "https://mockUrl.com",
+          "mockSubmitterKey",
+        );
       });
 
       it("Logs error", () => {
@@ -94,7 +101,10 @@ describe("getBiometricToken", () => {
           } as Response),
         ) as jest.Mock;
 
-        result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
+        result = await getBiometricToken(
+          "https://mockUrl.com",
+          "mockSubmitterKey",
+        );
       });
 
       it("Logs error", () => {
@@ -126,7 +136,10 @@ describe("getBiometricToken", () => {
         } as Response),
       ) as jest.Mock;
 
-      result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
+      result = await getBiometricToken(
+        "https://mockUrl.com",
+        "mockSubmitterKey",
+      );
     });
 
     it("Logs success at debug level", () => {
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
@@ -15,7 +15,7 @@ describe("getBiometricToken", () => {
 
   describe("On every call", () => {
     beforeEach(async () => {
-      async function httpRequestOverrideMockResponse() {
+      async function mockSendHttpRequestResponse() {
         return {
           statusCode: 200,
           body: "mockBody",
@@ -28,7 +28,7 @@ describe("getBiometricToken", () => {
       result = await getBiometricToken(
         "https://mockUrl.com",
         "mockSubmitterKey",
-        httpRequestOverrideMockResponse,
+        mockSendHttpRequestResponse,
       );
     });
 
@@ -42,14 +42,14 @@ describe("getBiometricToken", () => {
 
   describe("Given an error is caught when requesting a biometric access token", () => {
     beforeEach(async () => {
-      async function httpRequestOverrideMockError() {
+      async function mockSendHttpRequestErrorResponse() {
         throw new Error("mockError");
       }
 
       result = await getBiometricToken(
         "https://mockUrl.com",
         "mockSubmitterKey",
-        httpRequestOverrideMockError as unknown as ISendHttpRequest,
+        mockSendHttpRequestErrorResponse as unknown as ISendHttpRequest,
       );
     });
 
@@ -68,7 +68,7 @@ describe("getBiometricToken", () => {
   describe("Given the response is invalid", () => {
     describe("Given response body is undefined", () => {
       beforeEach(async () => {
-        async function httpRequestOverrideMockBodyUndefined() {
+        async function mockSendHttpRequestResponseBodyUndefined() {
           return {
             statusCode: 200,
             body: undefined,
@@ -81,7 +81,7 @@ describe("getBiometricToken", () => {
         result = await getBiometricToken(
           "https://mockUrl.com",
           "mockSubmitterKey",
-          httpRequestOverrideMockBodyUndefined,
+          mockSendHttpRequestResponseBodyUndefined,
         );
       });
 
@@ -99,7 +99,7 @@ describe("getBiometricToken", () => {
 
     describe("Given response body cannot be parsed", () => {
       beforeEach(async () => {
-        async function httpRequestOverrideMockBodyInvalidJSON() {
+        async function mockSendHttpRequestResponseBodyInvalidJson() {
           return {
             statusCode: 200,
             body: "Invalid JSON",
@@ -112,7 +112,7 @@ describe("getBiometricToken", () => {
         result = await getBiometricToken(
           "https://mockUrl.com",
           "mockSubmitterKey",
-          httpRequestOverrideMockBodyInvalidJSON,
+          mockSendHttpRequestResponseBodyInvalidJson,
         );
       });
 
@@ -136,7 +136,7 @@ describe("getBiometricToken", () => {
         expires_in: 3600,
         token_type: "Bearer",
       });
-      async function sendHttpRequestOverrideMockValidResponse() {
+      async function mockSendHttpRequestValidResponse() {
         return {
           statusCode: 200,
           body: mockBody,
@@ -149,7 +149,7 @@ describe("getBiometricToken", () => {
       result = await getBiometricToken(
         "https://mockUrl.com",
         "mockSubmitterKey",
-        sendHttpRequestOverrideMockValidResponse,
+        mockSendHttpRequestValidResponse,
       );
     });
 
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
@@ -18,7 +18,7 @@ describe("getBiometricToken", () => {
         Promise.resolve(),
       ) as jest.Mock;
 
-      result = await getBiometricToken("mockUrl", "mockSubmitterKey");
+      result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
     });
 
     it("Logs network call attempt at debug level", () => {
@@ -35,7 +35,7 @@ describe("getBiometricToken", () => {
         Promise.reject(new Error("Unexpected network error")),
       ) as jest.Mock;
 
-      result = await getBiometricToken("mockUrl", "mockSubmitterKey");
+      result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
     });
 
     it("Logs error", () => {
@@ -68,7 +68,7 @@ describe("getBiometricToken", () => {
           } as unknown as Response),
         ) as jest.Mock;
 
-        result = await getBiometricToken("mockUrl", "mockSubmitterKey");
+        result = await getBiometricToken("https://mockUrl.com", "mockSubmitterKey");
       });
 
       it("Logs error", () => {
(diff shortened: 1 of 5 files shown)


backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
```diff
@@ -1,10 +1,143 @@
...
+
+    it("Returns an empty failure", () => {
+      expect(result).toEqual(emptyFailure());
+    });
+  });
+
+  describe("Given the response is invalid", () => {
+    describe("Given response is undefined", () => {
+      it("Returns an empty failure", async () => {
+        /// Write test
```
jmooney-dd, line 56: Not sure this is possible but check is in logic due to TS

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
```diff
@@ -1,10 +1,76 @@
...
+    headers,
+  };
+
+  let response;
+  try {
+    logger.debug(
+      LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT,
+      {
+        data: {
+          httpRequest,
```
jmooney-dd, line 36: This will contain the submitter key, is this okay to log?

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
jmooney-dd: Removing and will use optional chaining anyway so this isn't an explicit check as sonar said it was a code smell

backend-api/template.yaml
```diff
@@ -160,7 +160,7 @@ Mappings:
   EnvironmentVariables:
     dev:
       STSBASEURL: 'https://mob-sts-mock.review-b-async.dev.account.gov.uk'
-      ReadIdBaseUrl: 'https://api-mob-readid-mock.review-b-async.dev.account.gov.uk'
+      ReadIdBaseUrl: 'https://api-mob-readid-mock.review-b-async.dev.account.gov.uk/v2'
```
jmooney-dd, line 163: The ReadID stack output does not include v2, is it worth adding this in the output or creating another output to show this?

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
sandymay-dd: Good question. No, it's a secret and should be treated with caution.

backend-api/template.yaml
jmooney-dd: I wonder what the best strategy here will be.

If the URL isn't correct, our API tests for this lambda will fail so I think it should be whatever the value is at the time we want to merge.. or I could disable that api test in this PR 🤷  

Also conscious Tiernan is doing a template update and don't want to cause him more headaches with conflicts!

sandymay-dd: Please may you keep a line of comms with Kiki - https://github.com/govuk-one-login/mobile-id-check-readid-mock/pull/15/files is the PR that will introduce this change.

sandymay-dd: we're taking off api- from the URL

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
```diff
@@ -1,10 +1,76 @@
...
+) => {
+  const readIdUrl = `${url}/oauth/token?grant_type=client_credentials`;
+  const headers = {
+    "X-Innovalor-Authorization": submitterKey,
+    "Content-Type": "application/x-www-form-urlencoded",
+  };
+
+  const httpRequest = {
+    url: readIdUrl,
+    method: "POST" as HttpMethod,
```
chris-cooksley-gds, line 26: Micro nitpick: I assume TS isn't happy if this isn't type-asserted to be HttpMethod, even though this string is part of the union type. If an assertion must be made, I wonder if `as const` offers a marginal improvement over `as HttpMethod`, on the basis that we're only telling typescript to infer the narrowest type for "POST", rather than to disregard type information entirely. In practical terms, I guess Typescript would probably not complain about`'ARBITRARYSTRING' as HttpMethod`, but would complain about `'ARBITRARYSTRING' as const`, so we gain a tiny bit more type safety.

Disclaimer: haven't tried this out so disregard all of this if `as const` doesn't make the compiler happy.

backend-api/template.yaml
jmooney-dd: Thanks - will do. Will keep this convo open for now

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
jmooney-dd: I just tried it and you're completely right. Will change now, thanks!

jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
```diff
@@ -1,10 +1,150 @@
...
+  let consoleDebugSpy: jest.SpyInstance;
+  let consoleErrorSpy: jest.SpyInstance;
+  beforeEach(() => {
+    consoleDebugSpy = jest.spyOn(console, "debug");
+    consoleErrorSpy = jest.spyOn(console, "error");
+  });
+
+  describe("On every call", () => {
+    beforeEach(async () => {
+      global.fetch = jest.fn(() => Promise.resolve()) as jest.Mock;
```
chris-cooksley-gds, line 17: Nitpick/thinking out loud:

I'm torn on mocking fetch itself here, rather than sendHttpRequest.

On the one hand, I see the value of trying not to tie the unit tests too much to the particular way we've split the functions/interfaces, and doing this allows us to change the interface of sendHttpRequest without updating this test.

But on the other hand, this test now needs to know about the implementation details of sendHttpRequest, the point of which is to abstract away a common utility. So if we wanted to change that implementation (e.g. to use axios instead of fetch), we'd have to not only update its own unit tests but these as well. Less important, but it's also a bit more cumbersome to mock the global fetch than it is to dependency-inject a mock sendHttpRequest.

For me personally the lesser evil would therefore be keeping them isolated and using jest to mock sendHttpRequest. An easy way to do this would be a final parameter on getBiometricToken that defaults to the real sendHttpRequest but can be overridden to use any implementation (and therefore to use a mock implementation) of the ISendHttpRequest interface.

jmooney-dd: I considered mocking the sendHttpRequest function and couldn't quite decide what would be better.

My rationale for mocking fetch only here is that I don't see as much value in keeping the test isolated. The getBiometricToken function will use sendHttpRequest so in my mind I'm happy for it to in the test as well. In this approach I wanted to only mock what I had to which was fetch.

Would be good to have a conversation about this if possible? Not suggesting my explanation/approach should be used, but a discussion to understand the best solution would help me a lot

jmooney-dd: After discussion I will mock `sendHttpRequest`. It is a shared util and we don't need to mock fetch every time we use it. It is harder to mock fetch and would require more far more tests to cover every scenario. If we were to change to axios this would be quite a long process to update all of our mocks whereas it is less likely will will change the interface of `sendHttpRequest`. 

Will update shortly

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
```diff
@@ -32,13 +40,14 @@ describe("getBiometricToken", () => {
 
   describe("Given there is an error making network request", () => {
     beforeEach(async () => {
-      global.fetch = jest.fn(() =>
-        Promise.reject(new Error("Unexpected network error")),
-      ) as jest.Mock;
+      mockSendHttpRequest = jest
+        .fn()
+        .mockRejectedValue(new Error("Unexpected network error"));
```
chris-cooksley-gds, line 45: I think if there were a network error, this would resolve to a failure object (rather than reject).

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
```diff
@@ -1,28 +1,33 @@
...
+import {
+  sendHttpRequest,
+  SuccessfulHttpResponse,
+} from "../../services/http/sendHttpRequest";
 import { emptyFailure, Result, successResult } from "../../utils/result";
 
 export type GetBiometricToken = (
   url: string,
   submitterKey: string,
+  sendHttpRequestOverride?: () => Promise<SuccessfulHttpResponse>,
```
chris-cooksley-gds, line 12: We already have a type for this function - `ISendHttpRequest`

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
```diff
@@ -1,28 +1,33 @@
...
 ) => Promise<Result<string, void>>;
 
 export const getBiometricToken: GetBiometricToken = async (
   url: string,
   submitterKey: string,
+  sendHttpRequestOverride?: () => Promise<SuccessfulHttpResponse>,
```
chris-cooksley-gds, line 18: This could be a default argument: `makeHttpRequest = sendHttpRequest`. Then we can get rid of the nullish coalescing below.

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
jmooney-dd: I have updated the test wording and removed the jest mocks. I think it now covers all cases in a clearer way. Let me know what you think

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
jmooney-dd: Updated thanks!

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
jmooney-dd: Updated

backend-api/template.yaml
jmooney-dd: Will monitor Kiki's PR and close this comment.

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.test.ts
chris-cooksley-gds: Two thoughts on this:

(1) in my previous comment I hadn't realised that our version of `sendHttpRequest` throws an exception on failure, instead of using the result pattern, hence my confusion about the mock rejected value. I think that's fine for now - but we should definitely look to update to using the result pattern (i.e. success with http response if it succeeds, failure with an appropriate error if not - cf. the [original STS version of the function](https://github.com/govuk-one-login/sts-back/blob/main/backend-api/src/common/http/sendHttpRequest.ts#L13)), to align with our conventional error handling pattern and with what's being used elsewhere in the pod. If you were planning to raise a followup PR to add in backoff/jitter etc anyway, it might be an idea to do this there?

(2) What was the rationale for removing the jest mocks? Was there behaviour that jest wasn't able to mock? I ask mainly as we have a [tech debt item](https://govukverify.atlassian.net/browse/DCMAW-10846) ready to go to switch to jest mocks as the preferred solution wherever possible, so we should look to favour them going forward unless we need something bespoke.

jmooney-dd: Re point 1 - yes I will be raising a f-up PR so can do that here

Re point 2 - Partly after a conversation with Sandy where he mentioned he wasn't a fan of using them unless we had to... I think here we just need a function that returns a simple object, rather than creating a whole class with methods on it (which is what that tech debt item was created to avoid I think). If it's just a function that returns a simple object then what advantage would we get by using a jest mock?

chris-cooksley-gds: The main advantage that comes with using a jest mock is built in functionality for verifying the number, order and arguments of calls. Verifying the precise number and order is admittedly a bit more specialised (I'd normally only do that for things like retries, caching etc) - but checking that the expected dependency _was in fact called_, and with the correct arguments, is crucial IMO if we're using mocking in the first place, so that we would get a test failure if we passed through the wrong arguments (which currently I don't think we would? e.g. how do we know that we're passing the submitter key in the header?)

Tagging @sandymay-dd as it would be good to align on approach here. I think we should have a team discussion on unit testing in the near future so we can move toward firming up some standards and conventions (I'd been thinking Dev Time next sprint but could bring forward given the number of times it's come up recently).

sandymay-dd: Sorry I'm a little spun in a few directions this morning so will go with the majority consensus here. 

My preference is where we can validate areas of risk without the need to check our implementation logic, I'd rather that. 

I'd rather use jest mocks over custom classes where we have to validate the implementation logic (as you mention @chris-cooksley-gds, retries is a great example); I'm not anti jest mocks. I'm anti overusing jest mocks to overcome poor code structure.

I'm thinking through this out loud: is there a way we could make use of our ReadID mock for this and enhance the ReadID mock's `/token` endpoint to do this validation? I wonder whether it's something we should be doing anyway.

sandymay-dd: And yes, a unit testing session would be good to document conventions (e.g. moving our unit test setup into before each, and having each assertion in separate it blocks; our infra test approach for environment variables)

chris-cooksley-gds: I'll set something up. I think we've all been mulling around our approach to unit testing, mocking, etc, especially since watching that [video]([🚀 TDD, Where Did It All Go Wrong (Ian Cooper) - YouTube](https://www.youtube.com/watch?v=EZ05e7EMOLM)) - but I'm conscious that we haven't necessarily formalised a practical pattern for doing this and maintaining good test coverage and easy test maintainability in our codebase. Until we do that, I'd prefer that we continue with our existing approach (broadly - functions/classes as isolated units, mocked out in tests to avoid tight coupling) until we have a more fleshed out game plan to replace it with a new one.

On the ReadID mock thing - my main concern is that we'd be widening our feedback loop by deferring some key aspects of test coverage to API tests that aren't run automatically on PR and _probably_ aren't run manually on every commit in the way that unit tests likely are.

sandymay-dd: Absolutely @chris-cooksley-gds -> we're great at touch and feel/exploring what works for us as we go. 

I'd lean to this in general as well:

Until we do that, I'd prefer that we continue with our existing approach (broadly - functions/classes as isolated units, mocked out in tests to avoid tight coupling) until we have a more fleshed out game plan to replace it with a new one.

mthkbarber (approved):

PR #340, opened 2025-01-16
[DCMAW-10830] update session state in /async/biometricToken

chris-cooksley-gds:
​DCMAW-10830

## What changed

* Introduces new `updateSession` method on the DynamoDB adapter. 
  * This method is defined as part of a generic SessionRegistry interface. 
  * It expects an UpdateSessionOperation, and uses the methods on that object to get the appropriate update/condition expressions for the UpdateItemCommand to DynamoDB. This isolates the logic for different types of update, facilitating testing and keeping the adapter logic simple.
  * It logs the attempt/success of the call to DynamoDB at debug level, as per the new pattern.
  * It logs conditional check failures separately to other failures, to help us identify when a session doesn't exist or is in the wrong state.
* Defines a `BiometricTokenIssued` update operation. It requires the session in question to exist, to (implicitly) not have expired, and to be in the ASYNC_AUTH_SESSION_CREATED state. It updates the state to ASYNC_BIOMETRIC_TOKEN_ISSUED and adds the passed document type and opaque ID.
* Updates handler to generate opaque ID and update the session using the above pattern. Conditional check failures elicit an unauthorized response, server errors elicit a server error response.
* Adds necessary permission to update session to the Lambda execution role
* Updates API tests for /async/biometricToken happy path, as the sessionId of a valid, current session is now required.

## Why did it change
* Our user session follows a forward-only state machine. Updating the session state - in tandem with conditional checks made on the current session state - ensures that the user's document type and opaque ID are 'locked in' and cannot be changed on subsequent requests.

## Evidence

image
image


backend-api/docs/sessionStateMachine.md
chris-cooksley-gds, line 1: Questions: 
* is this documentation useful?
* should it reflect the end state, or only where we're at so far (with updates to be made as we go along)?

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -9,3 +11,9 @@ export const expectedSecurityHeaders = {
 };
 
 export const NOW_IN_MILLISECONDS: number = 1704110400000; // 2024-01-01 12:00:00.000
+
+export const mockInertSessionRegistry: SessionRegistry = {
```
chris-cooksley-gds, line 15: Defining this here as a baseline will become increasingly valuable as we add more methods to the session registry, to avoid having to fill in inert implementations of the methods that aren't under test to appease Typescript when creating mock implementations.

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -43,6 +50,6 @@ describe("POST /async/biometricToken", () => {
       expect(response.headers).toEqual(
         expect.objectContaining(expectedSecurityHeaders),
       );
-    });
+    }, 15000);
```
chris-cooksley-gds, line 53: On my machine, it often takes longer than 10 seconds to perform all of the network calls needed here.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -66,6 +75,19 @@ export async function lambdaHandlerConstructor(
     return serverErrorResponse;
   }
 
+  const opaqueId = generateOpaqueId();
```
jmooney-dd, line 78: Nit: This function is just a wrapper for the randomUUID func. Would it be better to just call that here, that way I know quickly that opaqueId is a UUID?

chris-cooksley-gds: I was thinking of the fact that opaque ID is a UUID as an implementation detail that I wanted to hide/abstract away from the handler. It just so happens that we currently can generate it with a quick one liner, but I want it to be easy to change to a different approach.

backend-api/docs/sessionStateMachine.md
jmooney-dd: i like state machine docs. I think I would prefer the docs to only describe where we're at so far with the understanding we should double check this is updated with future state logic changes in the future

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -114,6 +120,49 @@ export class DynamoDbAdapter {
...
+      data: updateExpressionDataToLog,
+    });
+    try {
+      await this.dynamoDbClient.send(
+        new UpdateItemCommand({
+          TableName: this.tableName,
+          Key: {
+            sessionId: { S: sessionId },
+          },
+          UpdateExpression: updateOperation.getDynamoDbUpdateExpression(),
```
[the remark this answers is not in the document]
jmooney-dd: That makes sense and defo doesn't feel over engineered so I'm happy with that this

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -114,6 +120,49 @@ export class DynamoDbAdapter {
...
 
+  async updateSession(
+    sessionId: string,
+    updateOperation: UpdateSessionOperation,
+  ): Promise<Result<void, UpdateSessionError>> {
+    const updateExpressionDataToLog = {
+      updateExpression: updateOperation.getDynamoDbUpdateExpression(),
+      conditionExpression: updateOperation.getDynamoDbConditionExpression(),
+    };
+    logger.debug(LogMessage.UPDATE_SESSION_ATTEMPT, {
```
jmooney-dd, line 131: The tiniest nit

Feels slightly nicer to have this logger.debug for the attempt inside the try block. Obviously makes no function difference, feel free to ignore

backend-api/src/functions/adapters/tests/dynamoDbAdapter.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          error: "Conditional check failed",
+          data: {
+            updateExpression: updateOperation.getDynamoDbUpdateExpression(),
+            conditionExpression:
+              updateOperation.getDynamoDbConditionExpression(),
+          },
+        });
+      });
+
+      it("Returns failure with conditional check failure error", () => {
```
jmooney-dd, line 82: Tiny nit, it doesn't actually return the failure here

```suggestion
      it("Returns errorResult with conditional check failure error", () => {
```

jmooney-dd: This would go for the other "Returns failure..." blocks in this file. Feel free to dismiss

jmooney-dd (dismissed):

backend-api/src/functions/adapters/dynamoDbAdapter.ts
chris-cooksley-gds: You're right (I didn't realise I'd left it outside the block, actually) - will update

backend-api/src/functions/adapters/tests/dynamoDbAdapter.test.ts
chris-cooksley-gds: That's just me being set in my ways with MPT nomenclature. I may keep this as is, though, because even though the helper function is still called `errorResult` (one day I'll look to get that changed), the type is (now) called Failure/FailureWithValue/EmptyFailure

backend-api/docs/sessionStateMachine.md
chris-cooksley-gds: I think that makes more sense, as long as we remember to keep them up to date. Have pared down to the current impl

jmooney-dd (approved):

kikidawson-gds (approved):

PR #344, opened 2025-01-17
DCMAW-10829: Enhances http request retry strategy with exponential backoff and full jitter     

jmooney-dd:
### What changed
- sendHttpRequest:
  - Updates retry policy with exponential backoff and full jitter
  - Updates interface to use Result pattern after discussion feedback [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/337#discussion_r1918082211)
- Updates functions that consume sendHttpRequest as interface has changed:
  - getBiometricToken
  - publicKeyGetter (I made 2 additional changes listed below here we can revert if we want as minimal change as possible here):
    - Updated func in publicKeyGetter to use the Result pattern instead of throwing errors some of the time [cb1e74c2c47ae5847b9ca6db848c01bebd040a33](https://github.com/govuk-one-login/mobile-id-check-async/pull/344/commits/cb1e74c2c47ae5847b9ca6db848c01bebd040a33)
    - Updated publicKeyGetter to handle parsing error explicitly [8b3852d001eccaae9b2175b4687a736ae41da8c3](https://github.com/govuk-one-login/mobile-id-check-async/pull/344/commits/8b3852d001eccaae9b2175b4687a736ae41da8c3)
    - Update error message when response body does not match the expected JWKS structure [f3e03f2f916fecf11db31fa26f3bb312f7c0ee2f](https://github.com/govuk-one-login/mobile-id-check-async/pull/344/commits/f3e03f2f916fecf11db31fa26f3bb312f7c0ee2f)

#### Exponential backoff
Reduces load on target server giving it time to recover while gradually increasing the interval between retries

#### Full jitter

Prevents synchronised retries which can overwhelm the target server by altering the delay between retries by a random amount

#### Status

Resolved - Akin has resolved the hotspot: https://gds.slack.com/archives/C060PANF03T/p1737366725133829

#### Summary

Sonar flagged that we are using a pseudorandom number generator and asks we check it's safe to use here. We are using it to calculate a jitter value.

[Link to hotspot](https://sonarcloud.io/project/security_hotspots?sinceLeakPeriod=true&status=SAFE&pullRequest=344&id=mobile-id-check-async)

## ​Evidence

Deployed stack and ran `api` tests - all passing

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
@@ -34,6 +34,9 @@ export const getBiometricToken: GetBiometricToken = async (
       "X-Innovalor-Authorization": "Secret value and cannot be logged",
     },
   };
+  const retryConfig = {
+    retryableStatusCodes: [429, 500, 502, 503, 504],
+  };
 
   logger.debug(
     LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT,
@@ -44,7 +47,7 @@ export const getBiometricToken: GetBiometricToken = async (
     },
   );
   const getBiometricTokenResult: Result<SuccessfulHttpResponse, HttpError> =
-    await sendHttpRequest(httpRequest);
+    await sendHttpRequest(httpRequest, retryConfig);
   if (getBiometricTokenResult.isError) {
     logger.error(
       LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE,
--- a/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
+++ b/backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
@@ -1,8 +1,10 @@
 import { logger } from "../../common/logging/logger";
 import { LogMessage } from "../../common/logging/LogMessage";
 import {
+  HttpError,
   ISendHttpRequest,
   sendHttpRequest as sendHttpRequestDefault,
+  SuccessfulHttpResponse,
 } from "../../services/http/sendHttpRequest";
 import { emptyFailure, Result, successResult } from "../../utils/result";
 
@@ -33,36 +35,38 @@ export const getBiometricToken: GetBiometricToken = async (
     },
   };
 
-  let response;
-  try {
-    logger.debug(
-      LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT,
-      {
-        data: {
-          httpRequest: httpRequestLogData,
-        },
+  logger.debug(
+    LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_ATTEMPT,
+    {
+      data: {
+        httpRequest: httpRequestLogData,
       },
-    );
-    response = await sendHttpRequest(httpRequest);
-  } catch (error) {
+    },
+  );
+  const getBiometricTokenResult: Result<SuccessfulHttpResponse, HttpError> =
+    await sendHttpRequest(httpRequest);
+  if (getBiometricTokenResult.isError) {
     logger.error(
       LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE,
       {
         data: {
-          error,
+          error: getBiometricTokenResult.value,
           httpRequest: httpRequestLogData,
         },
       },
     );
+
     return emptyFailure();
   }
 
-  if (response?.body == null) {
+  const getBiometricTokenResponse = getBiometricTokenResult.value;
+
+  if (getBiometricTokenResponse?.body == null) {
     logger.error(
       LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE,
       {
         data: {
-          response,
+          getBiometricTokenResponse,
           httpRequest: httpRequestLogData,
         },
       },
@@ -72,7 +76,7 @@ export const getBiometricToken: GetBiometricToken = async (
 
   let parsedBody;
   try {
-    parsedBody = JSON.parse(response.body);
+    parsedBody = JSON.parse(getBiometricTokenResponse.body);
   } catch (error) {
     logger.error(
       LogMessage.BIOMETRIC_TOKEN_GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE,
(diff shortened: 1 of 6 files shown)


backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -21,36 +23,55 @@ export const sendHttpRequest: ISendHttpRequest = async (
...
+async function retryWithExponentialBackoffAndFullJitter(
+  request: () => Promise<Result<SuccessfulHttpResponse, HttpError>>,
+  attempt: number,
+  baseDelayMillis: number,
 ) {
-  await wait(delayInMillis);
+  const exponentialDelayWithoutJitter =
+    Math.pow(2, attempt - 1) * baseDelayMillis;
+  const exponentialDelayWithFullJitter = Math.floor(
+    Math.random() * exponentialDelayWithoutJitter,
```
jmooney-dd, line 72: Security hotspot using this 'weak' cryptographic function. 

_"Make sure that using this pseudorandom number generator is safe here."_

[Hotspot link](https://sonarcloud.io/project/security_hotspots?id=mobile-id-check-async&pullRequest=344&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

Lifted from STS backend so assume this is fine, we're just using it to calc a httpRequest delay, nothing sensitive.

jmooney-dd: Raised with Akin to review this hotspot: https://gds.slack.com/archives/C060PANF03T/p1737366725133829

backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -21,36 +23,55 @@ export const sendHttpRequest: ISendHttpRequest = async (
         headers,
         body,
       });
+
+      if (!response.ok && attempt < maxAttempts) {
```
chris-cooksley-gds, line 27: The ticket indicates that we should make 3 attempts following a 429 or 5XX response - but I think this implementation will retry anything that isn't 2xx (including non-retryable 4xx errors like 400/401, and I think even 3XX redirects).

I wonder if we should adopt the retryable status codes logic from the STS implementation, which would allow us to retry only what we want, and align the approach across the pod?

backend-api/src/functions/services/http/tests/sendHttpRequest.test.ts
```diff
@@ -76,6 +80,37 @@ describe("Send HTTP request", () => {
   });
 
   describe("Retry policy", () => {
+    describe("Given the status code is non-retryable and not 2xx", () => {
```
chris-cooksley-gds, line 83: Should we also have a test (or set of tests) for when the status code is retryable (i.e. 503, given the config above)? We could repurpose the 500 error test above for this.

backend-api/src/functions/services/http/sendHttpRequest.ts
jmooney-dd: Resolved

backend-api/src/functions/services/http/sendHttpRequest.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
```diff
@@ -32,37 +34,42 @@ export const getBiometricToken: GetBiometricToken = async (
       "X-Innovalor-Authorization": "Secret value and cannot be logged",
     },
   };
+  const retryConfig = {
+    retryableStatusCodes: [429, 500, 502, 503, 504],
```
jmooney-dd, line 38: Ticket asks for retry on 429 and 5XX. Obviously there are other 5XX errors but I'm not sure they are worth adding here. Do people agree or should I add every 5XX here? The ones I have not currently added:
- [501 Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501)
- [505 HTTP Version Not Supported](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/505)
- [506 Variant Also Negotiates](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/506)
- [507 Insufficient Storage](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/507)
- [508 Loop Detected](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/508) 
- [510 Not Extended](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/510)
- [511 Network Authentication Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/511)

chris-cooksley-gds: ^ thoughts @sandymay-dd ? I suppose it depends on how unpredictable the ReadID response codes could be?

backend-api/src/functions/services/http/tests/sendHttpRequest.test.ts
jmooney-dd: Updated

backend-api/src/functions/services/http/sendHttpRequest.ts
```diff
@@ -1,5 +1,8 @@
+import { errorResult, Result, successResult } from "../../utils/result";
+
 const DEFAULT_MAX_ATTEMPTS = 3;
 const DEFAULT_DELAY_IN_MILLIS = 100;
+const DEFAULT_RETRYABLE_STATUS_CODES = [408, 500, 502, 503, 504];
```
jmooney-dd, line 5: Copied the default codes over from STS. Do we think 429 should be here too? Unsure of the context behind why these are default, I don't suppose you remember @chris-cooksley-gds?

chris-cooksley-gds: I think we just informally cross-referenced what the AWS SDKs do with various recommendations on blog posts, stack overflows, etc to get a list to start with. No problem with adding 429 to this given that we're doing exponential backoff (so shouldn't be hammering any servers that are already getting too many requests).

backend-api/src/functions/asyncBiometricToken/getBiometricToken/getBiometricToken.ts
sandymay-dd: I'd err towards being over-zealous and going for all 5XX responses due to the lack of ReadID documentation, exactly as you mention @chris-cooksley-gds.

If we find there's a 5XX status code we _shouldn't_ be mapping, we'll find out in our 5XX error alarms and can adjust accordingly.

sandymay-dd: (I'm assuming we're logging the status codes for error scenarios here)

jmooney-dd: I have updated. I wonder if a 501 is too clear to not include in this list? No point retrying if whatever we're hitting isn't active/ready?

chris-cooksley-gds (approved):

PR #349, opened 2025-01-21
[DCMAW-10831] /async/biometricToken returns 200 with biometric access token and opaque ID

chris-cooksley-gds:
​DCMAW-10831

### What changed
* Updates biometric token handler to return 200 response with biometric access token from vendor, and generated opaque ID
* Updates associated tests
* Updates OpenAPI spec to reflect latest response fields

### Why did it change
* The mobile app will use the access token and opaque ID to start a biometric session with the vendor.

## Evidence

image

### DIFF
--- a/backend-api/openApiSpecs/async-public-spec.yaml
+++ b/backend-api/openApiSpecs/async-public-spec.yaml
@@ -236,16 +236,22 @@ paths:
                 success:
                   value:
                     {
-                      "accessToken": "string"
+                      "accessToken": "string",
+                      "opaqueId": "6ec96ea7-941c-4967-9fcf-94fc9b717a22"
                     }
               schema:
                 type: object
-                description: access token depending on passport, DL or BRP selection to access the third-party biometric check endpoints
                 required:
                   - accessToken
+                  - opaqueId
                 properties:
                   accessToken:
                     type: string
+                    description: access token depending on passport, DL or BRP selection to access the third-party biometric check endpoints
+                  opaqueId:
+                    type: string
+                    format: uuid
+                    description: generated token to be sent to the vendor when starting a new biometric session to facilitate fraud detection
         '400':
           description: Bad request
           content:
@@ -301,7 +307,6 @@ paths:
         uri:
           Fn::Sub: "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AsyncBiometricTokenFunction}:live/invocations"
         type: "aws_proxy"
-
   /async/finishBiometricSession:
     post:
       summary: App signifies to backend that user interaction for biometric session is complete
--- a/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
@@ -328,8 +328,8 @@ describe("Async Biometric Token", () => {
         headers: expectedSecurityHeaders,
         statusCode: 200,
         body: JSON.stringify({
-          access_token: "mockBiometricToken",
-          opaque_id: "mock_opaque_id",
+          accessToken: "mockBiometricToken",
+          opaqueId: "mock_opaque_id",
         }),
       });
     });
--- a/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
+++ b/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
@@ -1,5 +1,5 @@
 import { expect } from "@jest/globals";
-import "../testUtils/matchers";
+import "../../../tests/testUtils/matchers";
 import "dotenv/config";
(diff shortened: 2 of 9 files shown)


backend-api/openApiSpecs/async-public-spec.yaml
```diff
@@ -236,16 +236,22 @@ paths:
...
                 required:
                   - accessToken
+                  - opaqueId
                 properties:
                   accessToken:
                     type: string
+                    description: access token depending on passport, DL or BRP selection to access the third-party biometric check endpoints
+                  opaqueId:
+                    type: string
+                    format: uuid
```
jmooney-dd, line 253: Is it worth noting it's version 4 of uuid here? Actually not sure if that's overkill so feel free to resolve

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -323,11 +323,14 @@ describe("Async Biometric Token", () => {
...
-    it("Returns 501 Not Implemented response", async () => {
+    it("Returns 200 response with biometric access token and opaque ID", async () => {
       expect(result).toStrictEqual({
         headers: expectedSecurityHeaders,
-        statusCode: 501,
-        body: JSON.stringify({ error: "Not Implemented" }),
+        statusCode: 200,
+        body: JSON.stringify({
+          accessToken: "mockBiometricToken",
+          opaqueId: "mock_opaque_id",
```
jmooney-dd, line 332: I think in other unit tests we've not mocked UUID but instead asserted that is is in the shape of a UUID. Unsure how you feel about the value of that here or not?

jmooney-dd (dismissed):

backend-api/openApiSpecs/async-public-spec.yaml
chris-cooksley-gds: I don't know if that's something that you can specify with OpenAPI. Will probably keep as uuid as that's what's on the TD.

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
chris-cooksley-gds: I slightly prefer the approach I've gone with (but happy to be convinced). My main reasoning is that it verifies that we're returning a _random_ UUID - rather than just anything in the UUID format (i.e. if we only verify the shape, our tests would still pass if we used a hardcoded placeholder UUID). It's also slightly less cumbersome to implement, IMO.

sandymay-dd: I'm um-in and ah-ing.

In my head, this uuid is a sensitive value and we want to make sure we use a UUID rather than another algorithm for generating a random string.

Our acceptance criteria in the original penetration fix TD specifies this should be a UUID. I do wonder if it makes sense (potentially in our API tests) to validate the shape. This would also serve as potentially clearer documentation in our tests?

On the other hand, I don't think we're ever going to update the lambda to generate the value in a different format. 

tl;dr on the fence. Happy for you to go with your approach, @chris-cooksley-gds. Just playing devil's advocate.

chris-cooksley-gds: I'd been contemplating doing it in the api tests anyway (rather than just checking that it's a string). I'll keep this as-is and update the API test to check the shape.

backend-api/tests/testUtils/matchers.ts
```diff
@@ -36,12 +36,34 @@ function deepEquals(subject: unknown, target: unknown): boolean {
   return JSON.stringify(subject) === JSON.stringify(target);
 }
 
+const toBeValidUuid = function toBeValidUuid(candidate: unknown) {
```
jmooney-dd, line 39: What is the value of this wrapper opposed to just using the isValidUuid function you've created in the test?

jmooney-dd (dismissed):

backend-api/tests/testUtils/matchers.ts
chris-cooksley-gds: Just being able to use it as a Jest matcher, really. It's a nice bit of sugar to be able to say `expect(value).toBeValidUuid()` or, as we're doing, using an embedded 'asymmetric matcher' to check that a specific field of JSON is a valid UUID. Rather than pulling out the field and checking the value of a boolean function. 

I probably wouldn't have bothered if there wasn't a read-made implementation to copy-paste over.

jmooney-dd: Also is this just a different syntax I've not seen before or could this line be

```suggestion
const toBeValidUuid = (candidate: unknown) {
```

Ignore me if I'm just missing something obvious

sandymay-dd (dismissed):

backend-api/tests/testUtils/matchers.ts
chris-cooksley-gds: Good spot, not sure why this was done in the original implementation but as far as I can tell there's no functional difference in this case. Changed.

jmooney-dd (approved):

PR #380, opened 2025-02-07
DCMAW-10938: Writes DCMAW_ASYNC_CRI_5XXERROR event in async/biometricToken - error exchanging tokens and non conditional check errors updating session

jmooney-dd:
#### Backend
- Writes `DCMAW_ASYNC_CRI_5XXERROR` event in bometricToken when:
  - Any error during token token exchange with vendor
  - Errors when updating session unless it is a conditional check exception failure (in this case we write the `4XX` equivalent event)
  - FYI - when event fails to write, lambda will return `500` internal server error, however in the scenarios where these events are being written, the lambda was already going to return this same response
- Adds `DCMAW_ASYNC_CRI_5XXERROR` into `EventService` types and adds relevant tests

#### test-resources

- Added `DCMAW_ASYNC_CRI_5XXERROR` event to list of allowed TxMA event names so lambda could accept it and put it correctly into the event table in dynamo
- Did the same for `DCMAW_ASYNC_CRI_4XXERROR`. Was not done previously as this commit happened just before the dequeue lambda started pushing the events to the events table

### Why did it change

Audit requirements - see linked ticket for more information

### When reviewing
- TxMA event [specification](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3210018843/Specification+of+Messages+To+TXMA+for+Fraud#DCMAW_ASYNC_CRI_4XXERROR-and-DCMAW_ASYNC_CRI_5XXERROR)

The following fields have not been included in this event as they will be covered in this [ticket](https://govukverify.atlassian.net/browse/DCMAW-10989):
- `user.ipAddress`
- `restricted.device_information.encoded`

The following field has not been included as this check is re opaqueId mismatch which will be done in a lambda further in the flow
- `extensions.suspected_fraud_signal`

## Evidence

Manual testing

Forced the lambda to exit at the points detailed above and checked the event table to ensure event is successfully written:

Screenshot 2025-02-10 at 11 20 11

## Checklists


sts-mock/src/functions/dequeue/getEvent.ts
jmooney-dd, line 1: I don't think individual events are tested, just one from each, hence I have not added a test for this change. Shout if I have misunderstood

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -423,6 +470,53 @@ describe("Async Biometric Token", () => {
...
+        });
+
+        it("Logs the error", async () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            message: "ERROR_WRITING_AUDIT_EVENT",
+            function_arn: "arn:12345", // example field to verify that context has been added
+          });
+        });
+
+        it("Returns 500 Internal Server Error ", async () => {
```
jmooney-dd, line 498: Duplicate will remove

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -239,6 +239,53 @@ describe("Async Biometric Token", () => {
...
+      });
+
+      it("Logs the error", async () => {
+        expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+          message: "ERROR_WRITING_AUDIT_EVENT",
+          function_arn: "arn:12345", // example field to verify that context has been added
+        });
+      });
+
+      it("Returns 500 Internal Server Error ", async () => {
```
jmooney-dd, line 267: Duplicate - will remove

jmooney-dd: Removed

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jmooney-dd: Removed

sts-mock/src/functions/dequeue/getEvent.ts
jmooney-dd, line 1: I don't think individual events are tested, just one from each group of allowed TxMA event names, hence I have not added a test for this change. Shout if I have misunderstood

jmooney-dd:
Could I get people's opinion on test structure. 

Context
With 4XX events, they obvious exit the lambda with a 4xx response code but when they fail to write the 4xx event, the lambda will exit out with a 5xx code. This is how those tests are structured

```
      describe("Given foo", () => {
        describe("Given DCMAW_ASYNC_CRI_4XXERROR event fails to write to TxMA", () => {
          it("Logs the error", async () => {
          });

          it("Returns 500 Internal Server Error ", async () => {
          });
        });

        it("Writes DCMAW_ASYNC_CRI_4XXERROR event to TxMA", () => {
        });

        it("Returns 401 Unauthorized", () => {
        });
      });
```

Question
When 5xx errors, they are exiting the lambda with a 5xx code already, so when they fail to write the 5xx event, all that really happens is we log that. So the structure of the tests is a bit different since whether the event writing was ok or not, the lambda is always exiting with a 5xx code. So the test structure can either be:

(As is but happy to change if people prefer)

```
    describe("When bar", () => {
      describe("Given DCMAW_ASYNC_CRI_5XXERROR event fails to write to TxMA", () => {
        it("Logs the error", async () => {
        });
      });

      describe("Given DCMAW_ASYNC_CRI_5XXERROR event successfully writes to TxMA", () => {
        it("Writes DCMAW_ASYNC_CRI_5XXERROR event to TxMA", () => {
        });
      });

      it("Returns 500 Internal Server Error", () => {
      });
    });
```

OR

This is more in the style of the 4xx test at the top but we're testing for the same 500 twice

```
    describe("When bar", () => {
      describe("Given DCMAW_ASYNC_CRI_5XXERROR event fails to write to TxMA", () => {
        it("Logs the error", async () => {
        });

        it("Returns 500 Internal Server Error", () => {
        });
      });

      it("Writes DCMAW_ASYNC_CRI_5XXERROR event to TxMA", () => {
      });

      it("Returns 500 Internal Server Error", () => {
      });
    });
```

sts-mock/src/functions/dequeue/getEvent.ts
jmooney-dd: Confirmed fine

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -207,6 +213,29 @@ async function handleUpdateSessionSessionNotFound(
   return unauthorizedResponse("invalid_session", "Session not found");
 }
 
+async function handleWrite5xxEventInternalServerError(
```
sandymay-dd, line 216: suggestion: It sounds a bit like we're handling the error when the 5xx fails to send. Perhaps this could simply be `handleInternalServerError`

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -239,7 +239,54 @@ describe("Async Biometric Token", () => {
...
+          body: JSON.stringify({
+            error: "server_error",
+            error_description: "Internal Server Error",
+          }),
+          headers: expectedSecurityHeaders,
+        });
+      });
+    });
+
+    it("Writes DCMAW_ASYNC_CRI_5XXERROR event to TxMA", () => {
```
sandymay-dd, line 279: nit: Are we ok to get a describe block in here?

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -423,6 +470,53 @@ describe("Async Biometric Token", () => {
...
+            body: JSON.stringify({
+              error: "server_error",
+              error_description: "Internal Server Error",
+            }),
+            headers: expectedSecurityHeaders,
+          });
+        });
+      });
+
+      it("Writes DCMAW_ASYNC_CRI_5XXERROR event to TxMA", () => {
```
sandymay-dd, line 510: nit: same here

jmooney-dd: Added

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts
```diff
@@ -217,18 +242,14 @@ async function handleUpdateSessionError(
   switch (updateSessionResult.value.errorType) {
     case UpdateSessionError.CONDITIONAL_CHECK_FAILURE:
       sessionAttributes = updateSessionResult.value.attributes;
-      return handleUpdateSessionConditionalCheckFailure(
+      return handleUnauthorizedConditionalCheckFailure(
```
sandymay-dd, line 245:
```suggestion
      return handleConditionalCheckFailure(
```

I don't think we should couple the output of the failure we're handling with the cause of the failure

jmooney-dd: Updated and the other one that's like this

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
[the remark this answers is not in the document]
sandymay-dd: I think we need a test that validates the business requirement (we send session data to TxMA). 

I'm nervous where we could introduce errors that aren't mitigated through Types/Interfaces

1. We add new optional parameters to the write TxMA functions
2. We forget to pass through values in these parameters from the lambda
3. We wouldn't add test coverage for this/notice if we accidentally remove fields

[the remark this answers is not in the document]
sandymay-dd: (hadn't seen your update before I sent this)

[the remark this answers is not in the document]
sandymay-dd: @chris-cooksley-gds -> another one I think that will be a good contender for talking about testing 👀 

i.e. _How do we balance testing across API/lambda unit tests/Event writer unit tests?_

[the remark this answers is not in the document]
jmooney-dd: Updated

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -307,13 +357,14 @@ describe("Async Biometric Token", () => {
...
-                eventName: "DCMAW_ASYNC_CRI_4XXERROR",
-              }),
-            ),
-          );
+          expect(mockWriteGenericEventSuccessResult).toBeCalledWith({
+            eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+            componentId: "mockIssuer",
+            getNowInMilliseconds: Date.now,
+            govukSigninJourneyId: undefined,
+            sessionId: "58f4281d-d988-49ce-9586-6ef70a2be0b4",
```
sandymay-dd, line 365: suggestion: I think we've created a variable for this we've used elsewhere, avoiding hardcoding it in the test file

jmooney-dd: Updated

sandymay-dd (approved):

PR #404, opened 2025-02-11
fix pr workflow error

kikidawson-gds:
​DCMAW-XXXXX

### What changed
Testing whether this stops the error we occasionally see below.

### Why did it change
We are seeing an intermittent error in the `sam validate --lint` step of the PR workflows for test-resources and backend-api.

```
[[W8003: Fn::Equals will always return true or false] ([False, True] will always return False) matched 77]
Error: Linting failed. At least one linting rule was matched to the provided template.
```

## Checklists

### DIFF
--- a/backend-api/infra/parent.yaml
+++ b/backend-api/infra/parent.yaml
@@ -36,6 +36,11 @@ Parameters:
     Type: String
     Default: false
 
+  NeverDeployNullResource:
+    Description: NullResource exists to satisfy linter. This parameter ensures it is never deployed.
+    Type: String
+    Default: true
+
 Mappings:
   StaticVariables:
     urls:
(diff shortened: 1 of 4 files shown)


backend-api/infra/parent.yaml
```diff
@@ -256,8 +261,9 @@ Conditions:
         - !Ref Environment
         - dev
     - !Equals [!Ref DeployAlarmsInDev, true]
+
   Never: !Equals
-    - true
+    - !Ref NeverDeployNullResource
```
sandymay-dd, line 266: I've also realised - I think we only do this if we needed to do a `rain fmt` - we could remove this null resource from the codebase

sandymay-dd: we probably do want to do a `rain fmt` though!

kikidawson-gds: `rain fmt` seems to work without the resources section

kikidawson-gds:
Fix covered in this PR --> https://github.com/govuk-one-login/mobile-id-check-async/pull/409#pullrequestreview-2617305958

PR #405, opened 2025-02-11
DCMAW-11245: Do not provision lambda reserved concurrency in the dev environment

jmooney-dd:
### What changed
- Updates lambdas in dev so they do not provision reserved concurrency
  - Uses `isDev` conditional to set `ReservedConcurrentExecutions` to `AWS::NoValue` when environment is `dev`. Please note:
    - You cannot set `AWS::NoValue` in `Mappings` so initially I just removed it from the dev mapping, however...
    - Due to how Cloudformation deals with validation, it meant I had to have a value in the dev mapping even though it will never be used. I have set that value to 0 and added a comment (FYI - 0 is NOT the same as `AWS::NoValue`, setting value to 0 would mean lambda cannot fire)

### Why did it change
Currently each of our lambdas in dev have a ReservedConcurrentExecutions of 15 meaning just for one backend async stack alone (not including test-resources) it requires a 90 reserved concurrent executions. This number will increase when we add more lambdas.

As we only have 1000 ReservedConcurrentExecutions for the whole async dev account, and there is an additional limit that we must always have 100 capacity unused it all times, this has lead to developers being blocked from deploying stacks. This change will alleviate the ReservedConcurrentExecutions limit issues.

#####  Main

Screenshot 2025-02-11 at 15 05 52

#####  My stack

Screenshot 2025-02-11 at 15 00 42

#### Ran API tests

All passing

### What I didn't do

I did not deploy a stack from main, and then update it with these changes as I could not deploy a dev stack due to the minimum ReservedConcurrentExecutions being reached. I am personally happy to let this fail in the pipeline and revert it if there's an issue here. Shout if anyone has an issue with this

## Checklists


backend-api/infra/parent.yaml
```diff
@@ -266,10 +270,10 @@ Globals:
...
     Timeout: 5
     MemorySize: 512
-    ReservedConcurrentExecutions: !FindInMap
-      - Lambda
-      - !Ref Environment
-      - ReservedConcurrentExecutions
+    ReservedConcurrentExecutions: !If
+      - isDev
+      - !Ref AWS::NoValue
+      - !FindInMap [ Lambda, !Ref Environment, ReservedConcurrentExecutions ]
```
jmooney-dd, line 276: Might update to the style rain did it

backend-api/infra/parent.yaml
```diff
@@ -106,7 +106,7 @@ Mappings:
 
   Lambda:
     dev:
-      ReservedConcurrentExecutions: 15
+      ReservedConcurrentExecutions: 0 # Dummy value for CloudFormation's validation behaviour - will not be used
```
sandymay-dd, line 109: question: why do we need this?

sandymay-dd: I see your comment, but I'm still confused - and likely missing some Cloudformation behaviour

jmooney-dd: I think when you use an !If condition, Cloudformation will try to validate both branches which caused an error because it couldn't see ReservedConcurrentExecutions for dev when I had it removed

backend-api/infra/parent.yaml
jmooney-dd: Updated

backend-api/infra/parent.yaml
jmooney-dd: I could change this to an empty string if that would be nicer? Or look into it further if this doesn't feel right

jmooney-dd: Or update the comment to be clearer! 😅

sandymay-dd: I'm a tool - thank you

sandymay-dd (approved):

backend-api/infra/parent.yaml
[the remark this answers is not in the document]
jmooney-dd: Thanks, updated

jmooney-dd: Updated comment

sandymay-dd (approved):

PR #426, opened 2025-02-18
DCMAW-11007: Add TxMA event happy path assertions to backend API test

jhumbert-dd:
​DCMAW-11007

### What changed
- Adds TxMA event assertions to happy path API tests for:
  - POST /async/token
  - POST /async/biometricToken
- Adds test functions to the backend-api directory
  - `getEventsApiInstance()`
  - `pollForEvents()`
  - Function to check if an event was created in the last 60 seconds

##### If the function that checks an event's timestamp cannot find an event created in the last 60 seconds, an error is thrown:

Screenshot 2025-02-20 at 12 12 19


backend-api/tests/api-tests/utils/apiInstance.ts
```diff
@@ -60,7 +60,15 @@ function getStsMockInstance() {
   return getInstance(apiUrl);
 }
 
+function getEventsApiInstance() {
+  const apiUrl = process.env.EVENTS_API_URL;
```
jhumbert-dd, line 64: I think this env var needs to be created by the `generate_env_file.sh` script

backend-api/tests/api-tests/utils/apiTestHelpers.ts
```diff
@@ -90,3 +91,104 @@ export async function getValidSessionId(): Promise<string | null> {
...
+      pkPrefix: partitionKey,
+      skPrefix: sortKeyPrefix,
+    },
+  });
+
+  const events = response.data;
+  return Array.isArray(events) ? events : []; // If response is malformed, return empty array so polling can be retried
+}
+
+export function isEventLessThanOrEqualTo60SecondsOld(sortKey: string) {
```
jhumbert-dd, line 187: Is it better to get the `timestamp` from the `event` and create a type for `event` (which is currently `object`)

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -78,6 +87,26 @@ describe("POST /async/biometricToken", () => {
...
+        partitionKey: `SESSION#${sessionId}`,
+        sortKeyPrefix: `TXMA#EVENT_NAME#DCMAW_ASYNC_BIOMETRIC_TOKEN_ISSUED`,
+        numberOfEvents: 1,
+      });
+
+      const { pk, sk, event } = response[0];
+      const isEventCreationDateWithinTestTimeframe =
+        isEventLessThanOrEqualTo60SecondsOld(sk);
+
+      expect(isEventCreationDateWithinTestTimeframe).toBe(true);
```
jhumbert-dd, line 103: The reason for these lines can be found [here](https://govukverify.atlassian.net/browse/DCMAW-11007#:~:text=Note%2C%20this%20event,the%20same%20window.)

backend-api/tests/api-tests/utils/apiTestHelpers.ts
```diff
@@ -90,3 +91,104 @@ export async function getValidSessionId(): Promise<string | null> {
   );
   return activeSessionResponse.data["sessionId"] ?? null;
 }
+
+type EventResponse = {
+  pk: string;
+  sk: string;
+  event: object;
```
jhumbert-dd, line 98: Can/should `event` be typed? (relates to [comment below](https://github.com/govuk-one-login/mobile-id-check-async/pull/426/files#r1961422365))

backend-api/tests/api-tests/utils/apiTestHelpers.ts
sandymay-dd, line 187: What was the reason for choosing the timestamp from the dequeue lambda vs the timestamp from the lambda that sends the event?

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -78,6 +87,26 @@ describe("POST /async/biometricToken", () => {
...
+
+    it("Writes DCMAW_ASYNC_BIOMETRIC_TOKEN_ISSUED event to TxMA", async () => {
+      const response = await pollForEvents({
+        partitionKey: `SESSION#${sessionId}`,
+        sortKeyPrefix: `TXMA#EVENT_NAME#DCMAW_ASYNC_BIOMETRIC_TOKEN_ISSUED`,
+        numberOfEvents: 1,
+      });
+
+      const { pk, sk, event } = response[0];
+      const isEventCreationDateWithinTestTimeframe =
```
sandymay-dd, line 100: what's the rationale for needing to check the event age please?

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -78,6 +87,26 @@ describe("POST /async/biometricToken", () => {
...
+        isEventLessThanOrEqualTo60SecondsOld(sk);
+
+      expect(isEventCreationDateWithinTestTimeframe).toBe(true);
+      expect(pk).toEqual(`SESSION#${sessionId}`);
```
jmooney-dd, line 104: Nit: readability

It isn't immediately obvious to me what this assertion is checking. Is it the event we got back has the same sessionId so we know it's the correct event? 

Maybe it seems odd because we're testing a TxMA event but we're also checking a partitionKey.

I wonder if these tests can be made clearer adopting the new style of test structure we've been doing recently where you have the doing in the beforeEach, then we have an it for each assertion rather than multiple assertions here. This way you could use some english in an `it` block to detail what this assertion is for?

backend-api/tests/api-tests/utils/apiTestHelpers.ts
sandymay-dd: I would rather not maintain types for the TxMA events as we'll need to maintain them in two places (the source and test code)

backend-api/tests/api-tests/biometricToken.test.ts
jhumbert-dd: Apologies, this should only be in the /async/token tests. I will remove

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -144,6 +146,26 @@ describe.each(apis)(
...
+          partitionKey: `SESSION#UNKNOWN`,
+          sortKeyPrefix: `TXMA#EVENT_NAME#DCMAW_ASYNC_CLIENT_CREDENTIALS_TOKEN_ISSUED`,
+          numberOfEvents: 1,
+        });
+
+        const { pk, sk, event } = response[0];
+        const isEventCreationDateWithinTestTimeframe =
+          isEventLessThanOrEqualTo60SecondsOld(sk);
+
+        expect(isEventCreationDateWithinTestTimeframe).toBe(true);
```
jhumbert-dd, line 161: The reason for these lines can be found [here](https://govukverify.atlassian.net/browse/DCMAW-11007#:~:text=Note%2C%20this%20event,the%20same%20window.)

backend-api/tests/api-tests/biometricToken.test.ts
jhumbert-dd: Removed

backend-api/tests/api-tests/utils/apiTestHelpers.ts
jhumbert-dd: This is the same value as the event timestamp, which is populated by the lambda that sends the event.

backend-api/tests/api-tests/utils/apiTestHelpers.ts
```diff
@@ -90,3 +91,104 @@ export async function getValidSessionId(): Promise<string | null> {
...
+    },
+  });
+
+  const events = response.data;
+  return Array.isArray(events) ? events : []; // If response is malformed, return empty array so polling can be retried
+}
+
+export function isEventLessThanOrEqualTo60SecondsOld(sortKey: string) {
+  const eventTimestamp = parseInt(sortKey.split("#")[4]);
+  return eventTimestamp <= getTimeNowInSeconds();
```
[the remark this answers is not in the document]
jhumbert-dd: Fixed

backend-api/tests/api-tests/utils/apiTestHelpers.ts
jhumbert-dd: I've updated this to use the event timestamp (populated by the lambda that sends the event).

backend-api/tests/api-tests/biometricToken.test.ts
jhumbert-dd: Updated

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -140,9 +151,46 @@ describe.each(apis)(
...
+          eventsResponse = await pollForEvents({
+            partitionKey: `SESSION#UNKNOWN`,
+            sortKeyPrefix: `TXMA#EVENT_NAME#DCMAW_ASYNC_CLIENT_CREDENTIALS_TOKEN_ISSUED`,
+            numberOfEvents: 1,
+          });
+
+          ({ event } = eventsResponse[0]);
+        });
+
+        it("Occurred within the last 60 seconds", () => {
```
jmooney-dd, line 173: Suggestion only

I understand we are doing this because we need to make sure the event we are testing is the correct one, hence checking when it occurred but as this assertion is on the same level as the one below, it makes it seem like this is the same level of functionality we're testing.. rather than just a check

I wonder if instead of asserting anything we could do something like what we do in biometricToken test

```
      if (!sessionId)
        throw new Error(
          "Failed to get valid session ID to call biometricToken endpoint",
```

We could do this in the "Writes an event with the correct event_name" block
```
const isEventCreationDateWithinTestTimeframe =
  isEventLessThanOrEqualTo60SecondsOld(timestamp);

if (!isEventCreationDateWithinTestTimeframe) 
  throw new Error("something"
```

backend-api/tests/api-tests/utils/apiInstance.ts
jmooney-dd: Have these changes been tested using this  generate_env_file.sh script and the usual npm test method as well as running them pipeline style?

jhumbert-dd: Not yet 👍

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
jhumbert-dd: Nice - I've moved this check to the `beforeAll` block for these test.

backend-api/tests/api-tests/utils/apiInstance.ts
jmooney-dd: Happy to approve now once the above is done

jhumbert-dd: This has now been done

backend-api/tests/api-tests/utils/apiTestHelpers.ts
```diff
@@ -90,3 +91,106 @@ export async function getValidSessionId(): Promise<string | null> {
...
+    return Date.now();
+  }
+
+  function calculateExponentialBackoff(attempts: number) {
+    return Math.min(2 ** attempts * INITIAL_DELAY_MILLIS, MAX_BACKOFF_MILLIS);
+  }
+
+  const POLLING_DURATION_MILLIS = 40000; // maximum time to poll API
+  const MAX_BACKOFF_MILLIS = 10000; // maximum wait time between API calls
+  const INITIAL_DELAY_MILLIS = 500; // initial wait time before calling API
```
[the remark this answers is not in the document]
jhumbert-dd: I've removed the first one. I think the other two give a little bit more context than the names themselves. I have a really strong opinion about this one so happy to remove if preferred

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -65,19 +74,49 @@ describe("POST /async/biometricToken", () => {
...
         opaqueId: expect.toBeValidUuid(),
       });
-      expect(response.headers).toEqual(
+      expect(biometricTokenResponse.headers).toEqual(
         expect.objectContaining(expectedSecurityHeaders),
       );
-    }, 15000);
+    }, 5000);
+
+    describe("Writing to TxMA", () => {
```
jmooney-dd, line 94: nitpick: Do we need this describe block?

We could have the contents of beforeAll within this describe block, in the above beforeAll and then we could just have the three it statements

```
it("Writes an event with the correct session ID", () => {

it("Writes an event with the correct event_name", () => {

it("Returns a 200 response with biometric access token and opaque ID", async () => {
```
Entirely preference, will approve regardless of this

jmooney-dd (dismissed):

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
```diff
@@ -140,9 +149,38 @@ describe.each(apis)(
...
         expect(payload).toHaveProperty("scope", "dcmaw.session.async_create");
-        expect(response.data).toHaveProperty("expires_in", 3600);
-        expect(response.data).toHaveProperty("token_type", "Bearer");
-        expect(response.status).toBe(200);
+        expect(tokenResponse.data).toHaveProperty("expires_in", 3600);
+        expect(tokenResponse.data).toHaveProperty("token_type", "Bearer");
+        expect(tokenResponse.status).toBe(200);
+      });
+
+      describe("Writing to TxMA", () => {
```
jmooney-dd, line 157: Same point re my other nitpick re the test structure

backend-api/tests/api-tests/biometricToken.test.ts
jmooney-dd: Now I write it out, I do prefer your describe block, but we could have it in this order:

```
describe("Given there is a valid request", () => {

  describe("Writing to TxMA", () => {
    it("Writes a TxMA event with the correct session ID", () => {

    it("Writes a TxMA event with the correct event_name", () => {
  })

  it("Returns a 200 response with biometric access token and opaque ID", () => {
```

jhumbert-dd: Updated

backend-api/tests/api-tests/clientCredentialsGrantFlow.test.ts
jhumbert-dd: Updated

jmooney-dd (approved):

PR #429, opened 2025-02-20
DCMAW-11109: use latest version of cf template

tmulberry-gds:
​DCMAW-11109

### What changed
Lamba now uses the Session Table to validate that the supplied sessionId exists and meets certain validation criteria.  
If valid, it will update the state and save the biometricSessionId into the table.  
Added  TxMA events

### Why did it change
To validate the sessionId to be valid

### Evidence
image
image

## Checklists


backend-api/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "license": "MIT",
   "scripts": {
-    "deploy-backend-to-dev": "cd ../helper-scripts && ./deploy_backend.sh",
+    "deploy:dev": "cd ../helper-scripts && ./deploy_backend.sh",
```
jhumbert-dd, line 8: Nice 👌

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
```diff
@@ -100,6 +174,254 @@ describe("Async Finish Biometric Session", () => {
...
+          expect(result.statusCode).toBe(500);
+        });
+      });
+    });
+
+    describe("When session fails conditional check", () => {
+      describe("and session is expired", () => {
+        const expiredSessionAttributes = {
+          ...validBiometricSessionFinishedAttributes,
+          createdAt: Date.now() - 61 * 60 * 1000,
```
jmooney-dd, line 248: suggestion: Use faketimers and epoch format for createdAt

- We typically use `jest`'s fake timers in unit tests that rely on any sort of time validation, and then provide an invalid time before our mocked fake time, rather than relying on the real time and going backwards.
- I think using epoch (can't remember if it's secs or ms) might be better here as this is the actual format it's in and I don't think `Date.now() - 61 * 60 * 1000` drastically improves readability. We normally use the epoch time and then leave a comment next to it with a human date timestamp to say what the epoch time converts to

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
```diff
@@ -100,6 +174,254 @@ describe("Async Finish Biometric Session", () => {
     });
   });
 
+  describe("Session update scenarios", () => {
+    describe("When session not found", () => {
+      describe("and audit event succeeds", () => {
```
jmooney-dd, line 179: Suggestion: update order to test the fail writing scenario first

Non blocking but to be consistent, all other lambda handler tests follow a similar pattern with TxMA events. See the other lambda tests files for a proper example but it could be something like:

```
Given foo happens => {
  Given writing to Txma fails => {
    it returns 500 => {
  it writes event returns bar => {
```

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
jmooney-dd: This would go for other instances where we've done this but I won't comment on each of them

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -14,6 +17,142 @@ import {
...
+    await eventService.writeBiometricSessionFinishedEvent({
+      eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+      sub: sessionAttributes.subjectIdentifier,
+      sessionId: sessionAttributes.sessionId,
+      govukSigninJourneyId: sessionAttributes.govukSigninJourneyId,
+      componentId: issuer,
+      getNowInMilliseconds: Date.now,
+      transactionId: biometricSessionId,
+      extensions:
+        sessionAge > 60 * 60 * 1000
```
jmooney-dd, line 48: suggestion (non-blocking): Const this expression

You also use this expression below when checking if we should return a forbidden response, so we could do something like

const isSessionExpired = sessionAge > 60 * 60 * 1000

Then use this here and when we check for forbidden response

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -14,6 +17,142 @@ import {
...
+    });
+
+  if (writeEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: { auditEventName: "DCMAW_ASYNC_CRI_4XXERROR" },
+    });
+    return serverErrorResponse;
+  }
+
+  if (sessionAge > 60 * 60 * 1000) {
```
jmooney-dd, line 62: See this [comment](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965282103)

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -24,13 +163,43 @@ export async function lambdaHandlerConstructor(
   logger.addContext(context);
   logger.info(LogMessage.FINISH_BIOMETRIC_SESSION_STARTED);
 
-  const validateRequestBodyResult = validateRequestBody(event.body);
-  if (validateRequestBodyResult.isError) {
-    const errorMessage = validateRequestBodyResult.value.errorMessage;
+  const configResult = getFinishBiometricSessionConfig(dependencies.env);
+  if (configResult.isError) {
+    return serverErrorResponse;
+  }
```
jmooney-dd, line 169:
```suggestion
  const configResult = getFinishBiometricSessionConfig(dependencies.env);
  if (configResult.isError) {
    return serverErrorResponse;
  }
  
  const config = configResult.value
```

This is what we do in other lambdas. You can then access each envVar by doing `config.envVar` rather than doing `configResult.value.envVar`

I also add a comment where this can then be utilised

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -24,13 +163,43 @@ export async function lambdaHandlerConstructor(
...
-    return badRequestResponse("invalid_request", errorMessage);
+    return badRequestResponse(
+      "invalid_request",
+      validateResult.value.errorMessage,
+    );
+  }
+
+  const { sessionId, biometricSessionId } = validateResult.value;
+  const eventService = dependencies.getEventService(
+    configResult.value.TXMA_SQS,
```
jmooney-dd, line 184: See [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965289209)

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -24,13 +163,43 @@ export async function lambdaHandlerConstructor(
...
+  );
+  const sessionRegistry = dependencies.getSessionRegistry(
+    configResult.value.SESSION_TABLE_NAME,
```
jmooney-dd, line 187: See [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965289209)

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -24,13 +163,43 @@ export async function lambdaHandlerConstructor(
...
+    new BiometricSessionFinished(biometricSessionId),
+  );
+
+  if (updateResult.isError) {
+    return handleUpdateSessionError(
+      updateResult,
+      eventService,
+      sessionId,
+      biometricSessionId,
+      configResult.value.ISSUER,
```
jmooney-dd, line 201: See [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965289209)

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -94,6 +94,14 @@ export class LogMessage implements LogAttributes {
       "MOBILE_ASYNC_FINISH_BIOMETRIC_SESSION_REQUEST_BODY_INVALID",
       "The incoming request body was missing or invalid.",
     );
+  static readonly FINISH_BIOMETRIC_SESSION_INVALID_CONFIG = new LogMessage(
+    "MOBILE_ASYNC_FINISH_BIOMETRIC_SESSION_INVALID_CONFIG",
+    "One or more required environment variables were missing or invalid.",
+  );
+  static readonly FINISH_BIOMETRIC_SESSION_ERROR = new LogMessage(
```
jmooney-dd, line 101: issue: This log message isn't being used

Please can we add failure logs in the lambda where appropriate. We do have a common log message above for when audit events fail to write. If you need another log message for a different failure please may we make it specific to the failure, rather than just any error in this lambda

backend-api/tests/api-tests/finishBiometricSession.test.ts
```diff
@@ -1,37 +1,136 @@
...
+  //     expect(response.status).toBe(401);
+  //     expect(response.data).toStrictEqual({
+  //       error: "invalid_session",
+  //       error_description: "Session in invalid state",
+  //     });
+  //     expect(response.headers).toEqual(
+  //       expect.objectContaining(expectedSecurityHeaders),
+  //     );
+  //   });
+  // });
```
jmooney-dd, line 74: todo: please remove commented out tests

backend-api/tests/api-tests/finishBiometricSession.test.ts
```diff
@@ -1,37 +1,136 @@
...
+  //       },
+  //     );
+
+  //     expect(response.status).toBe(501);
+  //     expect(response.data).toStrictEqual({ error: "Not Implemented" });
+  //     expect(response.headers).toEqual(
+  //       expect.objectContaining(expectedSecurityHeaders),
+  //     );
+  //   });
+  // });
```
jmooney-dd, line 135: todo: please remove commented out tests if they are one of the scenarios we cannot legitimately test

backend-api/src/functions/common/session/updateOperations/sessionAttributes/sessionAttributes.ts
```diff
@@ -48,6 +49,18 @@ export const getBiometricTokenIssuedSessionAttributes = (
   return successResult(sessionAttributes);
 };
 
+export const getBiometricSessionFinishedSessionAttributes = (
```
jmooney-dd, line 52: todo: Please add a test for this, see `sessionAttributes.test.ts`. You can cover most of your assertions with a helper function that is already present in that file

backend-api/package.json
jmooney-dd: todo: This script is referenced in `backend-api/README.md`, please could you update

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  getSessionAttributesFromDynamoDbItem(
+    item: Record<string, AttributeValue> | undefined,
+    options?: {
+      operationFailed?: boolean;
+    },
+  ): Result<SessionAttributes, void> {
+    if (!item) {
+      return getBaseSessionAttributes(item);
```
jmooney-dd, line 40:
```suggestion
      return getBiometricTokenIssuedSessionAttributes(item);
```

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    if (options?.operationFailed) {
+      return getBaseSessionAttributes(item);
```
jmooney-dd, line 44:
```suggestion
      return getBiometricTokenIssuedSessionAttributes(item);
```

backend-api/package.json
jmooney-dd, line 8: todo: This script is referenced in backend-api/README.md, please could you update

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
tmulberry-gds: Is there a reason we're changing this one? currently it matches that in biometricTokenIssued

backend-api/src/functions/common/session/updateOperations/sessionAttributes/sessionAttributes.ts
jmooney-dd: This should take your coverage from 87.9 to 100

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
tmulberry-gds: Makes sense! Actioned.

jmooney-dd (changes requested):
todo: Since we can't API test the error TxMA events, can you manually trigger them and evidence in the PR? Happy to pair if you need help with this!

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -14,6 +17,143 @@ import {
...
+      return handleInternalServerError(
+        eventService,
+        sessionId,
+        biometricSessionId,
+        issuer,
+      );
+  }
+}
 
 export async function lambdaHandlerConstructor(
```
jmooney-dd, line 158: suggestion: can we keep this func at the top of the file to be consistent with the other lambda handlers

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -14,6 +17,143 @@ import {
...
+}
+
+async function handleSessionNotFound(
+  eventService: IEventService,
+  sessionId: string,
+  biometricSessionId: string,
+  issuer: string,
+): Promise<APIGatewayProxyResult> {
+  const writeEventResult =
+    await eventService.writeBiometricSessionFinishedEvent({
```
jmooney-dd, line 77: See comment [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965492877)

backend-api/src/functions/services/events/eventService.ts
```diff
@@ -40,6 +42,13 @@ export class EventService implements IEventService {
     return await this.writeToSqs(txmaEvent);
   }
 
+  async writeBiometricSessionFinishedEvent(
```
jmooney-dd, line 45: See comment [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429/files#r1965492877) re the name and usage of this function. This would also apply to the private funcs this calls

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -14,6 +17,143 @@ import {
...
+  sessionAttributes: SessionAttributes,
+  biometricSessionId: string,
+  issuer: string,
+): Promise<APIGatewayProxyResult> {
+  const sessionAge = Date.now() - sessionAttributes.createdAt;
+
+  const isSessionExpired = sessionAge > 60 * 60 * 1000;
+
+  const writeEventResult =
+    await eventService.writeBiometricSessionFinishedEvent({
```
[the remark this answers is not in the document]
jmooney-dd: Un resolving this - I think something need to change about this function, whether we use the generic one or rename it as the event is 4xx, not biometricSessionFinish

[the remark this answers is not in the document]
jmooney-dd: Few other relating to this so will un resolve just for visibility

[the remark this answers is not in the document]
jmooney-dd: IMO - I would go for using the generic function and then would update the `BaseTxmaEvent` type to be something like

```
export interface BaseTxmaEvent {
  timestamp: number;
  event_timestamp_ms: number;
  event_name: EventNames;
  component_id: string;
  restricted?: RestrictedData;
  extensions?: <the-type>
}
```

I would take this approach as extensions can be common over multiple different events and there's nothing else different about your 4 and 5xx events.

You can then look at the eventService in main to have a look how we deal with the restricted field which as you can see above is also optional. Happy to chat through this/look at a different approach if you had one in mind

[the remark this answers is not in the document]
tmulberry-gds: Resolved it first as I am a bit cautious to making too many changes to the eventService in this ticket, but I agree this can be a genericEvent. I have made the changes

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -23,20 +35,179 @@ export async function lambdaHandlerConstructor(
...
+  const writeEventResult = await eventService.writeGenericEvent({
+    eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+    sessionId,
+    sub: undefined,
+    govukSigninJourneyId: undefined,
+    componentId: issuer,
+    getNowInMilliseconds: Date.now,
+    transactionId: biometricSessionId,
+    extensions: {
+      suspected_fraud_signal: "AUTH_SESSION_NOT_FOUND",
```
jmooney-dd, line 136: I can't see this in this auth signals doc so I don't think this is a fraud signal

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -23,20 +35,179 @@ export async function lambdaHandlerConstructor(
...
+  const writeEventResult = await eventService.writeGenericEvent({
+    eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+    sub: sessionAttributes.subjectIdentifier,
+    sessionId: sessionAttributes.sessionId,
+    govukSigninJourneyId: sessionAttributes.govukSigninJourneyId,
+    componentId: issuer,
+    getNowInMilliseconds: Date.now,
+    transactionId: biometricSessionId,
+    extensions: isSessionExpired
+      ? { suspected_fraud_signal: "AUTH_SESSION_TOO_OLD" }
```
jmooney-dd, line 100: question: is this the correct name for this fraud signal?

There is conflicting info:
- On your ticket it says "AUTH_SESSION_TOO_OLD" which is what you have
- On the doc that is linked to the ticket for you to refer to re these signals, it says for finishBiometricSession it says the fraud signal is "BIOMETRIC_SESSION_OLDER_THAN_AUTH_SESSION" whereas "AUTH_SESSION_TOO_OLD" is for userInfo. This doc was made for the v1 backend so that could be why there's a difference but it's not clear

@sandymay-dd do you know what's correct here?

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
sandymay-dd: FYI https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/4453171396/DCMAW-10030+Resolve+Auth+Bypass+exploit?search_id=08fecc50-4029-4577-a6ca-239c99bd9408 is the tech design that introduced these signals

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
```diff
@@ -20,21 +25,65 @@ describe("Async Finish Biometric Session", () => {
   let consoleErrorSpy: jest.SpyInstance;
   let result: APIGatewayProxyResult;
 
+  // Constants for epoch timestamps
+  const CURRENT_TIME = 1708531200000; // 2024-02-21T12:00:00.000Z
```
jmooney-dd, line 29: nickpick (non-blocking): wondering if there's a different name because when you look below where this value is used, you might assume it is the time right now

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
sandymay-dd: The one on the tech design please (and the one registered in the event catalogue)

sandymay-dd: I think there's a mistake in the tech design

sandymay-dd: Which signal do we use for V1 in finishBiometricSession and userInfo/v2 please?

For v1:

I think BIOMETRIC_SESSION_OLDER_THAN_AUTH_SESSION should be userInfo/v2 as it's about checking the vendor session date with the id check session date

AUTH_SESSION_TOO_OLD should be for finishBiometricSession as it's only checking the id check session was created in the past 60 mins

sandymay-dd: We have have made an error in v1. If we have, we should make sure the async (this PR) implementation is correct

sandymay-dd: summary:

I believe the auth bypass tech design is wrong. If it is wrong:
- We implemented v1 correctly (i.e. not following the tech design)
- We've implemented this correctly (i.e. not following the tech design)

jmooney-dd: The other fraud signal I commented on [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/429#discussion_r1969481969), should go

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -23,20 +35,176 @@ export async function lambdaHandlerConstructor(
...
+      : undefined,
```
jmooney-dd, line 101: nitpick (non-blocking): we could have a helper function for this block

Originally Chris provided this feedback to me for the restricted blocks in the eventService but more because it reduced duplication I think. This wouldn't reduce duplication here but it would remove the cumbersome expression

```suggestion
    extensions: getFraudSignals(isSessionExpired)
```

Happy for you to resolve tho if you don't feel there's value

jmooney-dd (dismissed):

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -23,20 +35,181 @@ export async function lambdaHandlerConstructor(
...
+  const sessionAge = Date.now() - sessionAttributes.createdAt;
+  const isSessionExpired = sessionAge > 60 * 60 * 1000;
+
+  function getFraudSignal(
+    expired: boolean,
+  ): Record<string, string> | undefined {
+    if (expired) {
+      return { suspected_fraud_signal: "AUTH_SESSION_TOO_OLD" };
+    }
+  }
```
jmooney-dd, line 97:
```suggestion
  function getFraudSignal(
    expired: boolean,
  ): Record<string, string> | undefined {
    if (!expired) {
      return undefined
    }
    
    return { suspected_fraud_signal: "AUTH_SESSION_TOO_OLD" };
  }
```

Purely preference, I prefer to see a declarative return of undefined but please resolve if you don't agree

jmooney-dd (dismissed):

jmooney-dd (approved):

backend-api/src/functions/services/events/types.ts
```diff
@@ -74,10 +83,16 @@ export interface BiometricTokenIssuedEvent extends BaseUserTxmaEvent {
   };
 }
 
+export interface BiometricSessionFinishedEvent extends BaseUserTxmaEvent {
+  event_name: GenericEventNames;
+  extensions?: Extensions;
+}
```
jmooney-dd, line 89: Can this go now?

backend-api/src/functions/services/events/types.ts
```diff
@@ -74,10 +83,16 @@ export interface BiometricTokenIssuedEvent extends BaseUserTxmaEvent {
...
+
 export type TxmaEvents =
   | GenericTxmaEvent
   | CredentialTokenIssuedEvent
-  | BiometricTokenIssuedEvent;
+  | BiometricTokenIssuedEvent
+  | BiometricSessionFinishedEvent;
```
jmooney-dd, line 95: And here

backend-api/tests/api-tests/finishBiometricSession.test.ts
jmooney-dd, line 1: We spoke last week with Chris about how we couldn't API test most/if not all of the TxMA events. Just checking this still the case?

If this is still the case then nothing further needed. If testing a scenario where a TxMA event would be written, we should assert on it in this file

tmulberry-gds: This is still the case, we need an expired sessionId, which needs to be present in dynamo otherwise it will throw a not found instead of an expired session

kikidawson-gds (approved):

jmooney-dd (approved):

backend-api/tests/api-tests/finishBiometricSession.test.ts
tmulberry-gds: same with the "happy path" scenario, since it required a sessionId to be present, otherise it will throw "not found"

sandymay-dd: On the expired sessionId - we have no way to test this at the moment, and I think that's not a priority. As long as it's been tested manually, which I believe it has.

PR #446, opened 2025-02-27
[DCMAW-11433] Log STARTED message and append lambdaVersion key in async/credential 

chris-cooksley-gds:
​DCMAW-11433

### What changed and why
* Use our new logging pattern for async/credential, for consistency and maintainability
* Log MOBILE_ASYNC_CREDENTIAL_STARTED at the beginning of processing to support canary alarm work
* Remove erroneous import advice from logging docs

## Evidence

image

## Checklists

- [X] There is a ticket raised for this PR that is present in the branch name
- [X] No PII data logged. [See guidance here](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3502407722/PII+Logging+Considerations)

[DCMAW-11433]: https://govukverify.atlassian.net/browse/DCMAW-11433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -191,16 +199,22 @@ export async function lambdaHandlerConstructor(
...
     });
     return serverErrorResponse;
   }
 
-  logger.log("COMPLETED");
+  logger.info(LogMessage.CREDENTIAL_COMPLETED);
   return sessionCreatedResponse(requestBody.sub);
 }
 
+function setupLogger(context: Context) {
```
jmooney-dd, line 212: suggestion (non-blocking): worth making this a common function so other lambdas can use?

jmooney-dd: Or we could make it common whenever we update lambdas to use it in future

chris-cooksley-gds: I was thinking the latter if at all (if at all because we may want more flexibility for different lambdas to be set up in different ways - although I suppose there would probably be some common core to extract)

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -17,6 +28,16 @@ export class LogMessage implements LogAttributes {
     "Failed to retrieve one or more secrets from SSM Parameter Store.",
   );
 
+  static readonly CREATE_SESSION_FAILURE = new LogMessage(
+    "MOBILE_ASYNC_CREATE_SESSION_FAILURE",
+    "Failed to create user session in DynamoDB.",
+  );
+
+  static readonly GET_ACTIVE_SESSION_FAILURE = new LogMessage(
```
jmooney-dd, line 36: suggestion: move to credential section of this file

I don't think it's shared

jmooney-dd (dismissed):

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -17,6 +28,16 @@ export class LogMessage implements LogAttributes {
     "Failed to retrieve one or more secrets from SSM Parameter Store.",
   );
 
+  static readonly CREATE_SESSION_FAILURE = new LogMessage(
```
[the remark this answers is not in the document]
chris-cooksley-gds: So I'd done this initially, but my rationale is that both this and the log below are, in theory, referring to generic operations on the session (creating/reading), much like the other shared logs for session updates. It just happens that they're only used in one lambda currently - but I don't think that means they're inherently active session logs, and ultimately I'd want to log them in the session registry, which is shared. What do you think?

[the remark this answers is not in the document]
jmooney-dd: > ultimately I'd want to log them in the session registry, which is shared

Had not thought about this! And makes complete sense to me

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
sandymay-dd: just for reference, we're about to roll this out across all lambdas in the repository. This bit of context may not have been made explicitly clear (sorry). We may then want to make this common, although it's something we can pick up tomorrow

backend-api/src/functions/common/logging/LogMessage.ts
jmooney-dd: Discussed [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/446#discussion_r1973959892)

jmooney-dd (approved):

PR #447, opened 2025-02-28
DCMAW-11433: Updates logger in async/token to follow new logging pattern, appends keys, STARTED log

jmooney-dd:
### What changed and why
* Use our new logging pattern for async/token, for consistency and maintainability
* Log MOBILE_ASYNC_TOKEN_STARTED at the beginning of processing to support canary alarm work
* Adds shared setupLogger function, used it in token, credential, biometricToken, finishBiometricSession, txmaEvent

### Testing

I did not feel this change warranted a deployment and API testing. If you would like me to do this, let me know and I will. Unit tests have been updated.

## Checklists


backend-api/src/functions/common/logging/setupLogger.ts
sandymay-dd, line 1: question: Did you consider writing unit tests for this function? I'd be very happy not to if I'm honest, but just curious

jmooney-dd: We don't test our response helper lambda functions as it's tested directly in the lambda. The only difference here is that this function isn't in the lambda file but it is essentially a synchronous helper that's shared. 

I think based on this, I would leave as is unless you can think of a good reason to change?

backend-api/src/functions/asyncToken/asyncTokenHandler.ts
```diff
@@ -34,15 +36,15 @@ export async function lambdaHandlerConstructor(
   // Ensure that request contains expected params
   const eventBodyResult = requestService.validateBody(event.body);
   if (eventBodyResult.isError) {
-    logger.log("INVALID_REQUEST", {
+    logger.error(LogMessage.TOKEN_REQUEST_BODY_INVALID, {
```
sandymay-dd, line 39: questions[not requesting implementation change]: Do you know why this message isn't a common one; i.e. it `REQUEST_BODY_INVALID` and dropping the `TOKEN_`

jmooney-dd: Not aware of a reason no. Looking at the other lambda messages re invalid body there is no bespoke information there. Maybe a good candidate for a small tech improvement

backend-api/src/functions/common/logging/setupLogger.ts
sandymay-dd: I'd prefer to leave it as-is, but was curious for your thinking. Thanks for sharing it

PR #448, opened 2025-02-28
DCMAW-11474: Add request body validation

jhumbert-dd:
​DCMAW-11474

### What changed
- Adds function for validating the `sessionId` and `eventName` sent in the request body

### Why did it change
This is part of the epic to add TxMA billing events.


backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,124 @@
...
+  } catch (error: unknown) {
+    return errorResult({
+      handleErrorResponse: getErrorResponseHandler({
+        errorMessage: `Request body could not be parsed as JSON. ${error}`,
+      }),
+    });
+  }
+  const { sessionId, eventName } = parsedBody;
+
+  if (sessionId == null) {
```
sandymay-dd, line 30: suggestion: make this logic common

I think this will likely be similar, if not the same, to the validation we implemented in the asyncBiometricToken handler. I think we should make sure we're consistent across the board.

Happy to hear views on this one

jhumbert-dd: Yeah, I like the sound of that. I'm happy to extract a function for sessionId validation that the `validateRequestBody` function uses?

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,124 @@
+import { APIGatewayProxyResult } from "aws-lambda";
+import { badRequestResponse } from "../../common/lambdaResponses";
+import { logger } from "../../common/logging/logger";
+import { LogMessage } from "../../common/logging/LogMessage";
+import { errorResult, Result, successResult } from "../../utils/result";
+
+export function validateRequestBody(
```
jmooney-dd, line 7: question: should this function validate the sessionId is also a uuidv4? The tech design says it expects it to this.

Example [here](https://github.com/govuk-one-login/mobile-id-check-async/blob/53f15799f61f0534f548375a7fb9a291b03f2ef1/backend-api/src/functions/asyncFinishBiometricSession/validateRequestBody/validateRequestBody.ts#L35)

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,124 @@
...
+  "DCMAW_ASYNC_HYBRID_BILLING_STARTED",
+  "DCMAW_ASYNC_IPROOV_BILLING_STARTED",
+  "DCMAW_ASYNC_READID_NFC_BILLING_STARTED",
+];
+
+export interface IHandleErrorResponse {
+  handleErrorResponse: () => APIGatewayProxyResult;
+}
+
+function getErrorResponseHandler({ errorMessage }: { errorMessage: string }) {
```
sandymay-dd, line 117: suggestion: I think the AWS references (APIGatewayProxyResult) should live closer to the lambda file where possible. I don't think that keeping this here reduces the lambda file? Will come back to you after lunch

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jmooney-dd: Following on from this, I the openApi spec currently says sessionId should be a string, should we also update this to say something like

```
      properties:
        sessionId:
          type: string
          format: uuid
          example: 11111111-1111-1111-1111-111111111111
```

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jmooney-dd, line 117: nickpick/question (non-blocking): does 'get' work here in the func name

The name getErrorResponseHandler seems like you're going to get it to use later, but this function does the handling as well.

jmooney-dd: Oh sorry I think I understood how the function/pattern worked.

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -3,23 +3,29 @@ import {
...
+  event: APIGatewayProxyEvent,
   context: Context,
 ): Promise<APIGatewayProxyResult> {
   setupLogger(context);
   logger.info(LogMessage.TXMA_EVENT_STARTED);
 
+  const validateRequestBodyResult = validateRequestBody(event.body);
+  if (validateRequestBodyResult.isError) {
+    return validateRequestBodyResult.value.handleErrorResponse();
+  }
```
jmooney-dd, line 27: nickpick (non-blocking): Although slightly shorter, I don't think this is necessarily clearer than the pattern the other lambdas use

I don't mind certain parts of it like the logging happening in the validator function but I think this reads clearer:

```
  if (validateRequestBodyResult.isError) {
    return badRequestResponse(
      "invalid_request",
      validateRequestBodyResult.value.errorMessage,
    );
  }
```

I think it's specifically `validateRequestBodyResult.value.handleErrorResponse()` that is the odd part to me.

If there's other stuff I'm not thinking about/you disagree, please say!

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jmooney-dd: Sorry didn't see this, I [commented](https://github.com/govuk-one-login/mobile-id-check-async/pull/448#discussion_r1975433268) similar but in the lambda file

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -38,16 +45,58 @@ describe("Async TxMA Event", () => {
...
+        const request = buildRequest({
+          body: JSON.stringify({
+            sessionId: mockSessionId,
+            eventName: "INVALID_EVENT_NAME",
+          }),
+        });
+        result = await lambdaHandlerConstructor(dependencies, request, context);
+      });
+
+      it("Logs the error", async () => {
```
jmooney-dd, line 71: suggestion: move this test to the validator test file

The logging is currently being done in the validator function/file so I think we should test it there. We always used to test all logs in lambda files because we only logged in the lambda handler but that's now changed. 

If you search for log "GET_BIOMETRIC_TOKEN_FROM_READID_FAILURE" you'll see an example of the test in the function it's used in

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export interface IAsyncTxmaEventRequestBody {
+  sessionId: string;
+  eventName: EventName;
+}
+
+type EventName = (typeof allowableEventNames)[number];
+
+const allowableEventNames = [
```
sandymay-dd, line 107: suggestion: remove `allowable` from the variable name.

I don't think it adds much and I think the fact we validate the event name implies that we only recognise certain events

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jhumbert-dd: This change has been reverted

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
```diff
@@ -20,21 +22,10 @@ export function validateRequestBody(
...
+  if (validateSessionIdResult.isError) {
     return errorResult({
-      errorMessage: `sessionId in request body is not of type string. sessionId: ${sessionId}`,
-    });
-  }
-
-  if (sessionId === "") {
-    return errorResult({
-      errorMessage: `sessionId in request body is an empty string.`,
+      errorMessage: validateSessionIdResult.value.errorMessage,
```
jmooney-dd, line 28: Can we not just return `validateSessionIdResult`?

backend-api/src/functions/asyncFinishBiometricSession/validateRequestBody/validateRequestBody.ts
```diff
@@ -20,21 +22,10 @@ export function validateRequestBody(
...
-  if (!isString(sessionId)) {
-    return errorResult({
-      errorMessage: `sessionId in request body is not of type string. sessionId: ${sessionId}`,
-    });
-  }
-
-  if (!isValidUUIDv4(sessionId)) {
-    return errorResult({
-      errorMessage: `sessionId in request body is not a valid v4 UUID. sessionId: ${sessionId}`,
+      errorMessage: validateSessionIdResult.value.errorMessage,
```
jmooney-dd, line 28: question: same question here, can we not just return `validateSessionIdResult`

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,76 @@
...
+    return errorResult({
+      errorMessage: `Request body could not be parsed as JSON. ${error}`,
+    });
+  }
+  const { sessionId, eventName } = parsedBody;
+
+  const validateSessionIdResult = validateSessionId(sessionId);
+  if (validateSessionIdResult.isError) {
+    return errorResult({
+      errorMessage: validateSessionIdResult.value.errorMessage,
```
jmooney-dd, line 27: question: same again here, could we just return `validateSessionIdResult`

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
```diff
@@ -0,0 +1,76 @@
...
+    return errorResult({
+      errorMessage: `eventName in request body is not of type string. eventName: ${eventName}`,
+    });
+  }
+
+  if (eventName === "") {
+    return errorResult({
+      errorMessage: `eventName in request body is an empty string.`,
+    });
+  }
```
jmooney-dd, line 47: question: do we need this check?

My thinking is we check it is defined and that it's a string, could we skip this check and go straight to

`if (!isEventName(eventName))`

jmooney-dd: I think the empty string would get caught here, what do you think? Completely non blocking comment

backend-api/src/functions/asyncBiometricToken/validateRequestBody/validateRequestBody.ts
jhumbert-dd: updated

backend-api/src/functions/asyncFinishBiometricSession/validateRequestBody/validateRequestBody.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jhumbert-dd: updated

backend-api/src/functions/common/request/validateSessionId/validateSessionId.ts
```diff
@@ -0,0 +1,34 @@
...
+    });
+  }
+
+  if (!isValidUUIDv4(sessionId)) {
+    return errorResult({
+      errorMessage: `sessionId in request body is not a valid v4 UUID. sessionId: ${sessionId}`,
+    });
+  }
+
+  return successResult({ sessionId });
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/validateRequestBody/validateRequestBody.ts
jhumbert-dd: I think you're right. I also think it's a little bit cleaner to have a separate error message for an empty string rather than the result of the `isEventName` function returning '`...is invalid. eventName: `' with an empty value. I don't mind removing this if preferred but will leave as is unless maybe @sandymay-dd has preference?

jmooney-dd (approved):

PR #453, opened 2025-03-03
DCMAW-11465: Update TxMA events to include the redirect_uri

tmulberry-gds:
​DCMAW-11465

### What changed
Updated the async TxMA events in the lambda code to include the redirectUri retrieved from the Session table, if the user is on a MAM journey. 

### Why did it change
To log redirect_uri on MAM journeys where it is available.

### Evidence
image

image

image

## Checklists


backend-api/src/functions/common/helpers/txmaExtensions.ts
sandymay-dd, line 1: heya - this feels quite complex for our use case. I think there are other scenarios where the value we send to TxMA may be undefined. One scenario is where the update expression to Dynamo fails with a conditional error and the session record does (or doesn't) exist. 

Example scenario where the record doesn't exist - we explicitly pass through `undefined`: https://github.com/govuk-one-login/mobile-id-check-async/blob/53f15799f61f0534f548375a7fb9a291b03f2ef1/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts#L234


Here's the Type where we mark fields as potentially undefined: https://github.com/govuk-one-login/mobile-id-check-async/blob/53f15799f61f0534f548375a7fb9a291b03f2ef1/backend-api/src/functions/services/events/types.ts#L10

sandymay-dd: The key bit of info worth knowing, which I suspect may be a piece of tacit knowledge, is that if you send fields as `undefined` TxMA drop them when displaying in Splunk.

sandymay-dd: James is particularly close to this as I think he implemented the first change where information may or may not exist, so worth checking in with him if any of this needs more context etc.

tmulberry-gds: I have removed the getTxMAExtensions functions

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
```diff
@@ -186,10 +185,13 @@ export async function lambdaHandlerConstructor(
     govukSigninJourneyId: requestBody.govuk_signin_journey_id,
     getNowInMilliseconds: Date.now,
     componentId: config.ISSUER,
+    extensions: { redirect_uri: requestBody.redirect_uri },
```
jmooney-dd, line 188: suggestion (non-blocking): you could just pass through rediriectUrl here without extensions

Rationale is that this function simply passed through all values we need for the event, through to the event writer. It is then the event writers job to build the event so you can then use the buildTxmaEvent function to add the extensions block

Wondered if there was any specific reason we did it this way?

backend-api/src/functions/services/events/types.ts
```diff
@@ -65,6 +66,7 @@ export interface CredentialTokenIssuedEventConfig extends BaseEventConfig {
 
 export interface BiometricTokenIssuedEventConfig extends BaseUserEventConfig {
   documentType: DocumentType;
+  extensions?: Extensions;
```
jmooney-dd, line 69:
```suggestion (non-blocking)
  extensions: Extensions | undefined;
```

With this event specific event we always pass through the redirect uri we just don't know if it's defined or not, so would it make more sense to have this as mandatory to pass it through

Just a thought, happy if you've got a better reason not to do this

jmooney-dd: See comment re passing extensions through in general as well, we could just pass redirectUri here instead of extensions

tiernanmulberry: I have it within the object itself to simplify the eventService itsels and to have an easier type system for extensions

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
tmulberry-gds: I have it within the object itself to simplify the eventService itsels and to have an easier type system for extensions

backend-api/src/functions/services/events/types.ts
tmulberry-gds: Updated this

backend-api/src/functions/asyncCredential/asyncCredentialHandler.ts
jmooney-dd: Although it is easier for the event service this way, IMO it's still the eventService's job to the build the event. 

Also, if you look at the way the we build the `user` block, we never pass that through. The same with `extensions` `documentType` for `biometricTokenIssued` event.

So based on responsibility and consistency, I would update to pass through. Happy to discuss further tho if you don't agree!

tmulberry-gds: I have updated the eventService and references

backend-api/src/functions/services/events/eventService.ts
```diff
@@ -72,7 +72,10 @@ export class EventService implements IEventService {
       event_name: eventConfig.eventName,
       component_id: eventConfig.componentId,
       restricted: this.getRestrictedData(eventConfig.txmaAuditEncoded),
-      extensions: eventConfig.extensions,
+      extensions: {
+        redirect_uri: eventConfig.redirect_uri,
+        suspected_fraud_signal: eventConfig.suspected_fraud_signal,
+      },
```
jmooney-dd, line 78: question: Is there a situation where both redirectUri and the fraud signal could be undefined? If yes - what would the txma event look like if this happened? Would it end up being `extensions: {}`?

suggestion if so: create a helper function to either return undefined if both are undefined, meaning extensions will be dropped entirely, or to return what the object should be

jmooney-dd:
Discussed sonar duplication - suggestion is to look at using interfaces for our lambda response handler functions as we've done [here](https://github.com/govuk-one-login/mobile-id-check-async/blob/53f15799f61f0534f548375a7fb9a291b03f2ef1/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts#L183). I have only briefly looked at the sonar duplication report tho so might be worth seeing what you think as well

jmooney-dd:
Happy with this PR now although won't approve just yet as I think once you fix the duplication stuff it'll dismiss the review

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -66,149 +65,148 @@ export async function lambdaHandlerConstructor(
...
     ipAddress: undefined,
     txmaAuditEncoded: undefined,
   });
 
   if (writeEventResult.isError) {
     logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
-      data: { auditEventName: "DCMAW_ASYNC_CRI_4XXERROR" },
+      data: { auditEventName: eventName },
     });
+    // Always return server error if audit event writing fails
```
jmooney-dd, line 129: nitpick: please can we remove this comment

As we write lots of events across the codebase, and we always return a 500 when it fails to write, I think it's okay not to explain that here

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -66,149 +65,148 @@ export async function lambdaHandlerConstructor(
...
-  }
-
-  return unauthorizedResponse("invalid_session", "Session in invalid state");
+interface SessionErrorData {
+  sessionId: string;
+  biometricSessionId: string;
+  issuer: string;
+  eventName: GenericEventNames;
+  suspected_fraud_signal?: string;
+  sub?: string;
```
jmooney-dd, line 86: question: same as previous questions re the optional param on an interface, is this always passed we just don't know if it's a string or undefined, or do we not pass it sometimes? Not loaded, haven't checked.

If the former I think it should be `string | undefined`

Same point for all these interfaces and params in then

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -66,149 +65,148 @@ export async function lambdaHandlerConstructor(
...
+      data: { auditEventName: eventName },
     });
+    // Always return server error if audit event writing fails
     return serverErrorResponse;
   }
 
-  return unauthorizedResponse("invalid_session", "Session not found");
+  if (errorCode === "expired_session") {
+    return forbiddenResponse(errorCode, errorMessage);
+  } else if (errorCode.startsWith("invalid_")) {
```
jmooney-dd, line 135: question: would it be possible to do what you did above and do something like errorCode === "invalid_and-the-rest-of-it>"

tmulberry-gds: Done

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
tmulberry-gds: Yeah sorry whoops, habit. Updated it now

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
tmulberry-gds: Removed

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -79,6 +83,30 @@ export async function lambdaHandlerConstructor(
   return notImplementedResponse;
 }
 
+function buildEventConfig(
+  config: Partial<GenericEventConfig> &
+    Pick<
+      GenericEventConfig,
+      "eventName" | "sessionId" | "componentId" | "transactionId"
+    >,
+  includeSessionData: boolean = false,
```
jmooney-dd, line 92: note/question (non-blocking): this is the line that sonar flags as untested. Sonar isn't blocking this here as it's still got > 80 percent coverage, and I don't think we need to test this part of this helper func, does anyone else?

jmooney-dd: We do test all exit routers so I think it's fine

backend-api/src/functions/asyncCredential/asyncCredentialHandler.test.ts
```diff
@@ -972,6 +973,22 @@ describe("Async Credential", () => {
               "DCMAW_ASYNC_CRI_START",
             );
 
+            //Check the event config contains redirect url
```
jmooney-dd, line 976: nitpick (non-blocking): is this comment needed?

jmooney-dd (dismissed):
Approved based on API tests re run as discussed. It might be worth also getting @sandymay-dd's eyes on the duplication fix solution although not sure there's much better I can think of right now

backend-api/src/functions/services/events/eventService.ts
sandymay-dd, line 1: question: What was the rationale for the changes here? I'm not following what it changes

tmulberry-gds: If none of the extensions are populated, it prevents an empty extensions:{}

backend-api/src/functions/asyncCredential/asyncCredentialHandler.test.ts
tmulberry-gds: removed

jmooney-dd (approved):

PR #457, opened 2025-03-05
DCMAW-11030:  Updates logger in async/activeSession to follow new logging pattern, appends keys, STARTED log 

jmooney-dd:
### What changed and why

* Use our new logging pattern for async/activeSession, for consistency and maintainability
* Adds context as lambda parameter
* Log `MOBILE_ASYNC_ACTIVE_STARTED` at the beginning of processing to support canary alarm work
* Updates `activeSessionHander.test.ts` structure to match with new lambda test strat
* Replaces function for activeSessionHandler to get config with function that matches modern strat
* Creates new log for when active session is not found in activeSession lambda

### What I want to change in a follow-up PR

Refactor LogMessages re where they can be shared and do not need a bespoke message for each lambda

#### Automated

I would not normally deploy and test a logging change like this but I have due to touching a few different parts of the code. 

API tests result: All passing

#### Manual

Started log

Screenshot 2025-03-05 at 14 53 11

## Checklists


backend-api/src/functions/asyncActiveSession/getActiveSessionConfig.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return envVarsResult;
+}
+
+const isStsUrlValidUrl = (env: NodeJS.ProcessEnv): boolean => {
+  // To satisfy typescript, this has already been checked in getRequiredEnvironmentVariables
+  if (!env.STS_BASE_URL) {
+    return false;
+  }
```
jmooney-dd, line 50: Sonar note: sonar has picked up on sub 100 percent coverage. It is noting THIS in activeSession's as not being tested. I have put this in to satisfy TS as it could be undefined but  this check is already done on line 24 and is tested.

Happy to update/change how people wish/if they want me to

jmooney-dd: It's still at 95.6 percent so no blocking sonar issues, just wanted to highlight

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -130,6 +130,38 @@ export class LogMessage implements LogAttributes {
...
+  static readonly ACTIVE_SESSION_JWE_DECRYPTION_ERROR = new LogMessage(
+    "MOBILE_ASYNC_ACTIVE_SESSION_JWE_DECRYPTION_ERROR",
+    "Error whilst decrypting JWE service token.",
+  );
+  static readonly ACTIVE_SESSION_SERVICE_TOKEN_VALIDATION_ERROR =
+    new LogMessage(
+      "MOBILE_ASYNC_ACTIVE_SESSION_SERVICE_TOKEN_VALIDATION_ERROR",
+      "Error whilst validating decrypted service token.",
+    );
+  static readonly ACTIVE_SESSION_ERROR_LOOKING_UP_SESSION_IN_DB =
```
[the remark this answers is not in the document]
jmooney-dd: I didn't know whether to reference dynamo or just say db for database either

backend-api/src/functions/services/clientRegistryService/clientRegistryService.ts
```diff
@@ -143,7 +144,7 @@ export class ClientRegistryService
         typeof registeredClient.issuer === "string" &&
         typeof registeredClient.salt === "string" &&
         typeof registeredClient.hashed_client_secret === "string" &&
-        this.isValidUrl(registeredClient.redirect_uri) &&
+        isValidUrl(registeredClient.redirect_uri) &&
         Object.keys(registeredClient).length === 5,
```
jhumbert-dd, line 148: Should this check be done first? I'm not sure it checks anything new if the checks before pass

jmooney-dd: context: The change here is because I've removed the private isValidUrl from this class, and created a shared function instead. So all I've done here is import that func and use it.

I don't particularly want to change the functionality of this service in this commit as I don't think they're related enough. 

Out of curiosity, what do you mean re this check being done first? Isn't it just checking the url we got from the registry is valid?

jhumbert-dd: Sorry, (btw this is a nitpick and non-blocking) my comment was for the last check `Object.keys(...`. I think the functionality / validation would not change, just the order would. But as this is not part of the PR, feel free to leave as is.

jmooney-dd: Ah to check the length is 5? Yeah I think if we did want to change that it shouldn't be part of this PR.

Should it be changed? If the length was 6 couldn't the above still pass and it would fail this check?

jhumbert-dd: Would it pass if the check is `.length === 5` and the length was `6` or vice versa?

jmooney-dd: Apologies I don't quite follow. All these checks are part of allPropertiesPresent are they not?

I mean to say that if `registeredClient` looked something like:

```
{
  client_id: "mockClientId",
  issuer: "mockIssuer",
  salt: "mockSalt",
  hashed_client_secret: "mockSecret",
  redirect_uri: "www.mockUrl.com",
  extraKey: "someExtraKey"
  }
```

Wouldn't all of the current checks pass until it got to the length check and it would see there is one too many keys?

jhumbert-dd: I see - I think if the `allPropertiesPresent` function needs all checks to pass then I don't think the order of the checks affect the functionality unless I'm missing something?

backend-api/src/functions/common/logging/LogMessage.ts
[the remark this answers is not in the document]
sandymay-dd: Is it this log?

static readonly GET_ACTIVE_SESSION_FAILURE = new LogMessage(
    "MOBILE_ASYNC_GET_ACTIVE_SESSION_FAILURE",
    "Error occurred while attempting to retrieve active session by subject ID.",
  );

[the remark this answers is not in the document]
sandymay-dd: earlier up in the file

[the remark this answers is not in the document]
jmooney-dd: Updated

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -1,24 +1,29 @@
...
 import { RequestService } from "./requestService/requestService";
 import {
   dependencies,
   IAsyncActiveSessionDependencies,
 } from "./handlerDependencies";
-import { ConfigService } from "./configService/configService";
 import { ErrorCategory } from "../utils/result";
+import { setupLogger } from "../common/logging/setupLogger";
+import { logger } from "../common/logging/logger";
+import { LogMessage } from "../common/logging/LogMessage";
```
tmulberry-gds, line 14: question and nitpick from me, is there a reason we import these as separat files instead of under the folder, i.e. `import { LogMessage, setupLogger, logger } from "../common/logging`

jmooney-dd: To be honest I didn't even know that was a thing!

jmooney-dd: I tried to do that and it's erroring out, is there something I'm missing?

tmulberry-gds: Do you have an index.ts within the folder that exports them?

jmooney-dd: I don't have an index.ts file no, do you?

jmooney-dd: Sorry I'm probably completely misunderstanding something here 😅

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
```diff
@@ -58,36 +60,34 @@ export async function lambdaHandlerConstructor(
...
   if (getActiveSessionResult.isError) {
-    logger.log("INTERNAL_SERVER_ERROR", {
+    logger.error(LogMessage.GET_ACTIVE_SESSION_FAILURE, {
       errorMessage: getActiveSessionResult.value.errorMessage,
     });
     return serverErrorResponse;
   }
   const session = getActiveSessionResult.value;
   if (session === null) {
+    logger.error(LogMessage.ACTIVE_SESSION_ACTIVE_SESSION_NOT_FOUND);
```
sandymay-dd, line 86: suggestion: logger.info

Our system hasn't errored here (in theory it's acting exactly as we want it to).

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
sandymay-dd: @tmulberry-gds can you please get this on the tech improvements confluence page as one for our usual triage process?

sandymay-dd (dismissed):

backend-api/src/functions/asyncActiveSession/asyncActiveSessionHandler.ts
jmooney-dd: Updatede

backend-api/src/functions/asyncActiveSession/getActiveSessionConfig.ts
jmooney-dd: It's because the way we check if it's a url is doing `new Url()` which can only take type string. So typescript needs this proved that it's a string and not undefined or it errors, hence this check. 

We have already checked it's there and string in getRequiredEnvironmentVariables so this check seems wasteful so I added the comment.

Happy to adapt the logic or comment if you have any suggestions. I could just do `return isValidUrl(env.STS_BASE_URL!);` but this feels more janky

backend-api/src/functions/asyncActiveSession/getActiveSessionConfig.ts
```diff
@@ -0,0 +1,53 @@
...
+    });
+    return emptyFailure();
+  }
+
+  return envVarsResult;
+}
+
+const isStsUrlValidUrl = (env: NodeJS.ProcessEnv): boolean => {
+  // To satisfy typescript, this has already been checked in getRequiredEnvironmentVariables
+  if (!env.STS_BASE_URL) {
```
sandymay-dd, line 48: question: why are we accessing this from .env? Can we use `envVarsResult.value.STS_BASE_URL`?

sandymay-dd: apologies this is for line 46. We know it's going to be a string based on the getRequiredEnvironmentVariables, so the input to this function can be `stsBaseUrl: string`?

backend-api/src/functions/asyncActiveSession/getActiveSessionConfig.ts
jmooney-dd: Implementing the solution on your other comment

backend-api/src/functions/asyncActiveSession/getActiveSessionConfig.ts
jmooney-dd: Updated

sandymay-dd (approved):

PR #469, opened 2025-03-10
DCMAW-11475: Add session validation to TxmaEvents Lambda

jhumbert-dd:
​DCMAW-11475

### What changed
- Adds `getSession` method to `DynamoDbAdapter` to retrieve and validate a session. Validation checks the session:
  - exists
  - is in the `ASYNC_BIOMETRIC_TOKEN_ISSUED` state
  - is less than or equal to `60` minutes old.
- Updates existing validation
  - Removes validation from `sessionAttributes` type guards.
  - Adds `validateSession` method to existing Dynamo operations for:
    - `BiometricTokenIssued.ts`
    - `BiometricSessionFinished.ts`
  - Refactors `sessionAttributes` type guards to reduce duplication

#### TxMA 4XX Error event written to SQS
Test data: `mockSessionId = "58f4281d-d988-49ce-9586-6ef70a2be0b4"`

Screenshot 2025-03-20 at 11 21 49

Screenshot 2025-03-20 at 11 20 36

### Next steps
- [x] Remove unnecessary tests if needed
- [ ] Refactor log and error data typing
- [x] Add test data to remove duplication from new tests
- [x] Extract session validation to common location

### Why did it change
This is part of the work to add new TxMA events to track the billable transactions during a user’s biometric verification journey.


backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -131,6 +141,55 @@ export class DynamoDbAdapter implements SessionRegistry {
     }
   }
 
+  async getSession(
+    sessionId: string,
+  ): Promise<Result<void, SessionRetrievalFailed>> {
+    logger.debug(LogMessage.GET_SESSION_ATTEMPT);
```
jmooney-dd, line 147: suggestion: can we move this to within the try block? I think this is typically how we do this elsewhere for network attempt stuff

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -235,4 +306,28 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+    item: Record<string, AttributeValue> | undefined,
+    options?: {
+      operationFailed: boolean;
+    },
+  ): Result<SessionAttributes, void> {
+    if (options?.operationFailed) {
+      return getBaseSessionAttributes(item);
+    } else {
+      return getBiometricTokenIssuedSessionAttributes(item);
+    }
```
jmooney-dd, line 331: though: should this decision making me done in the dynamo adapter or should we pass something through it, like we do with the update function, to decide which type of session attributes we get back?

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,6 +48,26 @@ export async function lambdaHandlerConstructor(
...
+    config.SESSION_TABLE_NAME,
+  );
+  const getSessionResult = await sessionRegistry.getSession(sessionId);
+  const eventService = dependencies.getEventService(config.TXMA_SQS);
+  const eventData = {
+    sessionId,
+    issuer: config.ISSUER,
+    ipAddress: getIpAddress(event),
+    txmaAuditEncoded: getHeader(event.headers, "Txma-Audit-Encoded"),
+  };
```
jmooney-dd, line 63:
```suggestion
  const { ipAddress, txmaAuditEncoded } = getAuditHeaders(event);
  const eventData = {
    sessionId,
    issuer: config.ISSUER,
    ipAddress,
    txmaAuditEncoded,
  };
```

I wouldn't commit from here as they'll be import changes

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -131,6 +141,55 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+    const { createdAt } = sessionAttributes;
+    if (this.isOlderThan60minutes(createdAt)) {
+      logger.error(LogMessage.GET_SESSION_SESSION_TOO_OLD);
+
+      return errorResult({
+        errorType: GetSessionError.SESSION_NOT_FOUND,
+      });
+    }
+
+    logger.error(LogMessage.GET_SESSION_SUCCESS);
```
jmooney-dd, line 188: issue: this is not an error

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -131,6 +141,55 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+
+    return emptySuccess();
```
jmooney-dd, line 190: issue: Do we not also need to check the state here? At the moment I can only see we have checked createdAt

jmooney-dd, line 190: issue: Do we not also need to check the state before we return success? At the moment I can only see we have checked createdAt

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -177,11 +236,10 @@ export class DynamoDbAdapter implements SessionRegistry {
         const getSessionAttributesOptions = {
           operationFailed: true,
         };
-        const getAttributesResult =
-          updateOperation.getSessionAttributesFromDynamoDbItem(
-            error.Item,
-            getSessionAttributesOptions,
-          );
+        const getAttributesResult = this.getSessionAttributesFromDynamoDbItem(
```
jmooney-dd, line 239: question: why are we changing this?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -205,8 +263,9 @@ export class DynamoDbAdapter implements SessionRegistry {
       }
     }
 
-    const getAttributesResult =
-      updateOperation.getSessionAttributesFromDynamoDbItem(response.Attributes);
+    const getAttributesResult = this.getSessionAttributesFromDynamoDbItem(
```
jmooney-dd, line 266: question: why are we changing this, should this not come from the updateOperation?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: That's done in the `getSessionAttributesFromDynamoDbItem` function

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: I wanted to make this a common function as this is now not an exclusive `updateOperation`. I think this might need to be part of a bigger refactor that might be too big to include in the changes for this PR

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: The reasoning for this is the same as this [comment](https://github.com/govuk-one-login/mobile-id-check-async/pull/469#discussion_r1987371714)

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
jhumbert-dd: Nice! Updated

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jmooney-dd: Your function does this:
```
    if (options?.operationFailed) {
      return getBaseSessionAttributes(item);
    } else {
      return getBiometricTokenIssuedSessionAttributes(item);
    }
```
    
But different lambdas use updateSession and ask for different shapes of attributes to what is above

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jmooney-dd: None functionally. Discussed it with Chris and we preferred it being in the try block as it resembles it properly being the attempt 🤷

backend-api/src/functions/adapters/dynamoDbAdapter.ts
```diff
@@ -131,6 +142,55 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+      this.getSessionAttributesFromDynamoDbItem(responseItem);
+    if (getSessionAttributesResult.isError) {
+      return this.handleGetSessionNotFoundError(
+        "Could not parse valid session attributes after successful get command",
+      );
+    }
+    const sessionAttributes = getSessionAttributesResult.value;
+
+    const { createdAt } = sessionAttributes;
+    if (this.isOlderThan60minutes(createdAt)) {
```
jmooney-dd, line 181: question: did you consider using a query command so you didn't have to do this validation here?

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: Updated to use the same pattern as the `updateSession` method

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: Updated

backend-api/src/functions/adapters/dynamoDbAdapter.ts
jhumbert-dd: Moved to validation to the Lambda after discussion.

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -90,8 +145,281 @@ describe("Async TxMA Event", () => {
...
+            headers: expectedSecurityHeaders,
+          });
+        });
+      });
+
+      it("Writes DCMAW_ASYNC_CRI_4XXERROR event to TxMA", () => {
+        expect(mockWriteGenericEventSuccessResult).toBeCalledWith({
+          eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+          componentId: "mockIssuer",
+          getNowInMilliseconds: expect.anything(),
```
jhumbert-dd, line 385: For some reason, using `Date.now` for this value does not match when using jest `useFakeTimers` and/or `setSystemTime`. Is there any other way to not use this workaround?

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxMAEvent.ts
```diff
@@ -0,0 +1,24 @@
...
+export class TxMAEvent implements GetSessionOperation {
+  private readonly sessionId: string;
+
+  constructor({ sessionId }: { sessionId: string }) {
+    this.sessionId = sessionId;
+  }
+
+  getDynamoDbKeyExpression() {
+    return { sessionId: marshall(this.sessionId) };
+  }
```
jmooney-dd, line 17: nitpick (non-blocking): we could just pass sessionId to the getSession func, I don't necessarily think it needs it's own func here

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -42,3 +87,129 @@ export const lambdaHandler = lambdaHandlerConstructor.bind(
...
+  return true;
+}
+
+function isOlderThan60minutes(createdAtInMilliseconds: number) {
+  const SIXTY_MINUTES_IN_MILLISECONDS = 3600000;
+  const validFrom = getTimeNowInMilliseconds() - SIXTY_MINUTES_IN_MILLISECONDS;
+
+  console.log("<<<<< IOT60M TIME NOW >>>>>", getTimeNowInMilliseconds());
+  console.log("<<<<< IOT60M CREATED AT >>>>>", createdAtInMilliseconds);
+  console.log("<<<<< IOT60M VALID FROM >>>>>", validFrom);
```
jmooney-dd, line 208: issue: assume these will be removed

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -42,3 +87,129 @@ export const lambdaHandler = lambdaHandlerConstructor.bind(
...
+}
+
+function validateSession(sessionAttributes: SessionAttributes): boolean {
+  const { sessionState, createdAt } = sessionAttributes;
+  console.log("<<<<< VS 1 >>>>>");
+  if (sessionState !== SessionState.BIOMETRIC_TOKEN_ISSUED) return false;
+  console.log("<<<<< VS 2 >>>>>");
+
+  if (isOlderThan60minutes(createdAt)) return false;
+  console.log("<<<<< VS 3 >>>>>");
```
jmooney-dd, line 197: issue: can we remove console.log

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -42,3 +87,129 @@ export const lambdaHandler = lambdaHandlerConstructor.bind(
...
+    txmaAuditEncoded,
+    ...genericTxmaEventData,
+  });
+}
+
+function validateSession(sessionAttributes: SessionAttributes): boolean {
+  const { sessionState, createdAt } = sessionAttributes;
+  console.log("<<<<< VS 1 >>>>>");
+  if (sessionState !== SessionState.BIOMETRIC_TOKEN_ISSUED) return false;
+  console.log("<<<<< VS 2 >>>>>");
```
jmooney-dd, line 194: issue: can we remove console.log

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
jmooney-dd, line 1: issue: there are console.logs still in this file, please can we remove

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +304,129 @@ describe("DynamoDbAdapter", () => {
...
+        expect(result).toEqual(
+          errorResult({
+            errorType: GetSessionError.SESSION_NOT_FOUND,
+          }),
+        );
+      });
+    });
+
+    describe("Given session was found", () => {
+      describe("Given invalid session attributes were returned in response", () => {
```
jmooney-dd, line 370: question: is this not the same test as the session not found test [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/469/files#diff-21922729a9b2f03203c4d24babee8687788eb5574f12ed141f79a70b7ec9670cR348)?

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +304,129 @@ describe("DynamoDbAdapter", () => {
...
+        });
+
+        beforeEach(async () => {
+          mockDynamoDbClient.on(GetItemCommand).resolves({
+            Item: validBiometricTokenIssuedSessionAttributesItem,
+          });
+          result = await sessionRegistry.getSession(getOperation);
+        });
+
+        it("Logs the failure", () => {
```
jmooney-dd, line 416:
```suggestion
        it("Logs the success", () => {
```

or something similar

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -56,6 +77,40 @@ describe("Async TxMA Event", () => {
...
+      (envVar: string) => {
+        beforeEach(async () => {
+          delete dependencies.env[envVar];
+          result = await lambdaHandlerConstructor(
+            dependencies,
+            validRequest,
+            context,
+          );
+        });
+        it("returns 500 Internal server error", async () => {
```
jmooney-dd, line 92:
```suggestion
        it("returns 500 Internal server error", () => {
```

backend-api/generate_env_file.sh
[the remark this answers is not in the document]
jhumbert-dd: Reverted

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
jmooney-dd: In biometricSessionFinished lambda unit tests we use fake timers but still call this with Date.now, having a look at what's different...

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
jmooney-dd: Discussed, I thought this was the lambda file, ignore me!

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
jmooney-dd: Discussed and we have a fix

jhumbert-dd: Resolved. Thanks @jmooney-dd

backend-api/tests/api-tests/txmaEvent.test.ts
```diff
@@ -44,9 +49,46 @@ describe("POST /async/txmaEvent", () => {
...
+
+  describe("Given the request is valid", () => {
+    let sessionId: string | null;
+    let response: AxiosResponse;
+
+    beforeAll(async () => {
+      const sub = randomUUID();
+      await createSessionForSub(sub);
+      sessionId = await getActiveSessionIdFromSub(sub);
+      if (!sessionId) throw new Error("Failed to get valid session ID");
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -56,6 +77,40 @@ describe("Async TxMA Event", () => {
...
+          expect(result).toStrictEqual({
+            statusCode: 500,
+            body: JSON.stringify({
+              error: "server_error",
+              error_description: "Internal Server Error",
+            }),
+            headers: expectedSecurityHeaders,
+          });
+        });
+        it("logs INVALID_CONFIG", async () => {
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxMAEvent.ts
jhumbert-dd: Updated following discussion

kikidawson-gds (approved):

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,6 +50,34 @@ export async function lambdaHandlerConstructor(
...
+    return handleGetSessionError(
+      eventService,
+      getSessionResult.value,
+      eventData,
+    );
+  }
+  const sessionAttributes = getSessionResult.value;
+
+  const isSessionValid = validateSession(sessionAttributes);
+  if (!isSessionValid) return handleSessionNotFound(eventService, eventData);
```
jmooney-dd, line 80:
```suggestion
  if (!validateSession(sessionAttributes)) return handleSessionNotFound(eventService, eventData);
```

Entirely non-blocking. My thought is we don't use the const anywhere else and the function name provides enough meaning imo, happy for you to resolve tho

jmooney-dd: This or the validateSession could also return a Result and then we follow the typical isError pattern

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -37,6 +37,31 @@ export class LogMessage implements LogAttributes {
...
+
+  static readonly GET_SESSION_SESSION_NOT_FOUND = new LogMessage(
+    "MOBILE_ASYNC_GET_SESSION_SESSION_NOT_FOUND",
+    "No session found when attempting to get the user session from DynamoDB.",
+  );
+
+  static readonly GET_SESSION_SESSION_TOO_OLD = new LogMessage(
+    "MOBILE_ASYNC_GET_SESSION_SESSION_TOO_OLD",
+    "Session was not created within the valid timeframe",
+  );
```
jmooney-dd, line 58: issue: we are not currently using this

question: If we do want to log this, should the name be different as the check is in the lambda not in the getSession call

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -131,6 +140,47 @@ export class DynamoDbAdapter implements SessionRegistry {
     }
   }
 
+  async getSession(
+    sessionId: string,
+    getOperation: GetSessionOperation,
+  ): Promise<Result<SessionAttributes, SessionRetrievalFailed>> {
+    let response;
+    try {
+      logger.debug(LogMessage.GET_SESSION_ATTEMPT);
```
jmooney-dd, line 149: suggestion (non-blocking): worth logging some extra data here? Both the sessionId and getOperation do not contained PII and therefore could be logged

If you do decide to log this, please ensure you assert on the whole log here as we always want to ensure no PII is getting logged in these debug logs

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -42,3 +88,117 @@ export const lambdaHandler = lambdaHandlerConstructor.bind(
...
+function isOlderThan60minutes(createdAtInMilliseconds: number) {
+  const SIXTY_MINUTES_IN_MILLISECONDS = 3600000;
+  const validFrom = getTimeNowInMilliseconds() - SIXTY_MINUTES_IN_MILLISECONDS;
+
+  return createdAtInMilliseconds < validFrom;
+}
+
+function getTimeNowInMilliseconds() {
+  return Math.floor(Date.now());
+}
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -58,3 +41,68 @@ export const validBiometricSessionFinishedAttributes = {
...
+export const mockSuccessfulSessionRegistry: SessionRegistry = {
+  updateSession: jest.fn().mockResolvedValue(
+    successResult({
+      attributes: validBiometricTokenIssuedSessionAttributesMobileApp,
+    }),
+  ),
+
+  getSession: jest.fn().mockResolvedValue(
+    successResult({
+      attributes: validBaseSessionAttributes,
```
jmooney-dd, line 69: suggestion: I think it would be more accurate to use validBiometricTokenIssuedSessionAttributes here

backend-api/tests/api-tests/txmaEvent.test.ts
jmooney-dd, line 1: Question: should we be testing and asserting on a TxMA event here?

backend-api/src/functions/common/logging/LogMessage.ts
jhumbert-dd: Updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
jhumbert-dd: updated

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -90,8 +148,277 @@ describe("Async TxMA Event", () => {
...
+  });
+
   describe("Given a valid request is made", () => {
     beforeEach(async () => {
+      dependencies.getSessionRegistry = () => ({
+        ...mockSuccessfulSessionRegistry,
+        getSession: jest.fn().mockResolvedValue(
+          successResult({
+            ...validBaseSessionAttributes,
+            sessionState: SessionState.BIOMETRIC_TOKEN_ISSUED,
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/testUtils/unitTestData.ts
jhumbert-dd: updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -131,6 +139,57 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+            queryOperation.getDynamoDbKeyConditionExpression(),
+        }),
+      );
+    } catch (error: unknown) {
+      const errorData = {
+        ExpressionAttributeValues:
+          queryOperation.getDynamoDbExpressionAttributeValues(sessionId),
+        KeyConditionExpression:
+          queryOperation.getDynamoDbKeyConditionExpression(),
+        error,
```
jmooney-dd, line 165: suggestion: log these in the attempt log and only error data on an error log

backend-api/tests/api-tests/txmaEvent.test.ts
jhumbert-dd: TxMA assertion has been added.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            .resolves(biometricTokenIssuedSessionAttributesSessionTooOldItem);
+          result = await sessionRegistry.getSession(
+            mockSessionId,
+            getOperation,
+          );
+        });
+
+        it("Logs the failure", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode: "MOBILE_ASYNC_GET_SESSION_SESSION_NOT_FOUND",
```
jhumbert-dd, line 450: Missing `session` field?

jhumbert-dd: Added

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.test.ts
jhumbert-dd, line 1: I think `sessionState` and `createdAt` validation assertions should be added to this file

backend-api/tests/api-tests/biometricToken.test.ts
```diff
@@ -98,7 +98,7 @@ describe("POST /async/biometricToken", () => {
         "/async/biometricToken",
         requestBody,
       );
-    }, 15000);
+    }, 25000);
```
jhumbert-dd, line 101: Increased following the rule of thumb of `5000` ms for each `await` (including nested ones) after having an intermittent failure.

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -90,8 +148,403 @@ describe("Async TxMA Event", () => {
...
+              error_description: "Session does not exist or in incorrect state",
+            }),
+            headers: expectedSecurityHeaders,
+          });
+        });
+      });
+    });
+  });
+
+  describe("Session validation", () => {
```
jhumbert-dd, line 334: I think this can be condensed into one test as the `dynamoAdapter` tests cover the conditional branching for session validation.

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.ts
sandymay-dd, line 1: suggestion: please may you look at updating the unit tests in sessionAttributes.test.ts for these changes?

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+          errorResult({
+            errorType: GetSessionError.SESSION_NOT_FOUND,
+            session: "Session not found",
+          }),
+        );
+      });
+    });
+
+    describe("Session validation", () => {
+      describe("Given the session is in the wrong state", () => {
```
sandymay-dd, line 380: prompt for tomorrow: how should we write unit tests for the UpdateSession function validation. Should we be testing every implementation of the UpdateSession operation within this adapter?

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
       });
     });
   });
+
+  describe("getSession", () => {
+    const getOperation: GetSessionOperation = new TxMAEvent();
+    let result: Result<void, SessionRetrievalFailed>;
```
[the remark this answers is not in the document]
jhumbert-dd: Yeah, I believe `void` translates to a success response without a value (`emptySuccess()`)

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+          }),
+        );
+      });
+    });
+
+    describe("Session validation", () => {
+      describe("Given the session is in the wrong state", () => {
+        const invalidBiometricTokenIssuedSessionAttributesWrongSessionState = {
+          ...validBiometricTokenIssuedSessionAttributes,
+          createdAt: 1704106860000, // 2024-01-01 11:01:00.000
```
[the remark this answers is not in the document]
jhumbert-dd: no reason - I'll remove this

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.ts
```diff
@@ -37,6 +36,14 @@ export const isBaseSessionAttributes = (
   return true;
 };
 
+export const isBaseSessionAttributes = (
+  item: Record<string, NativeAttributeValue>,
+): item is BaseSessionAttributes => {
+  if (!isCommonSessionAttributes(item)) return false;
+  if (typeof item.sessionState !== "string") return false;
+  return true;
```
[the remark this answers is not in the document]
jhumbert-dd: This was part of my working out of how type guard functions work and is leftover mistake. I've updated to include sessionState in the base attributes

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.test.ts
jhumbert-dd: My reason was because some validation was already being done here before this work. Validation has now been moved to another layer.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -1,29 +1,37 @@
...
+  GetItemCommand,
   ReturnValue,
   ReturnValuesOnConditionCheckFailure,
   UpdateItemCommand,
 } from "@aws-sdk/client-dynamodb";
+import { marshall, unmarshall } from "@aws-sdk/util-dynamodb";
+import { expect } from "@jest/globals";
 import { mockClient } from "aws-sdk-client-mock";
-import { errorResult, Result, successResult } from "../../../utils/result";
+import "dotenv/config";
```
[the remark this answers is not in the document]
jhumbert-dd: removed

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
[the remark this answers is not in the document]
jhumbert-dd: removed

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -235,4 +294,34 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+  ): FailureWithValue<SessionRetrievalFailedInternalServerError> {
+    logger.error(LogMessage.GET_SESSION_UNEXPECTED_FAILURE, {
+      error,
+    });
+    return errorResult({
+      errorType: GetSessionError.INTERNAL_SERVER_ERROR,
+    });
+  }
+
+  private handleGetSessionNotFoundError({
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.ts
jhumbert-dd: updated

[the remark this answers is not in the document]
jhumbert-dd: removed unused export declarations

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
       });
     });
   });
+
+  describe("getSession", () => {
+    const getOperation: GetSessionOperation = new TxMAEvent();
```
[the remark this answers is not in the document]
jhumbert-dd: Updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+
+    describe("Given there is an unexpected error retrieving the session", () => {
+      beforeEach(async () => {
+        mockDynamoDbClient.on(GetItemCommand).rejects("mock_error");
+        result = await sessionRegistry.getSession(mockSessionId, getOperation);
+      });
+
+      it("Logs the failure", () => {
+        expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+          messageCode: "MOBILE_ASYNC_GET_SESSION_UNEXPECTED_FAILURE",
```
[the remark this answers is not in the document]
jhumbert-dd: I wasn't able to test within the `toHaveBeenCalledWithLogFields` assertion so added separate ones below. It would be great to know if there's a better way to do this

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
jhumbert-dd: I think one test for attribute type validation and one test for session should be enough. I'll remove one of the session validation tests from each method

jhumbert-dd: updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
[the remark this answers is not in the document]
jhumbert-dd: @sandymay-dd If you're happy with the way I've asserted on the error I'll resolve this comment. cc @chris-cooksley-gds

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -58,3 +43,68 @@ export const validBiometricSessionFinishedAttributes = {
   sessionState: SessionState.BIOMETRIC_SESSION_FINISHED,
   biometricSessionId: mockBiometricSessionId,
 };
+
+export const validBiometricTokenIssuedSessionAttributesMobileApp = {
+  ...validBiometricSessionFinishedAttributes,
```
chris-cooksley-gds, line 48: Should this be using the BiometricTokenIssued attributes here?

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,13 +50,184 @@ export async function lambdaHandlerConstructor(
...
+
+  const auditEventName = "DCMAW_ASYNC_CRI_4XXERROR";
+  if (writeEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: { auditEventName },
+    });
+    return serverErrorResponse;
+  }
+
+  logger.error(LogMessage.TXMA_EVENT_SESSION_NOT_FOUND, {
```
chris-cooksley-gds, line 165: We'll already be logging `GET_SESSION_SESSION_NOT_FOUND` in the dynamo adapter when this happens, so I think we can omit this log message to avoid double-logging the same error.

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,13 +50,184 @@ export async function lambdaHandlerConstructor(
...
+  if (writeEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: {
+        auditEventName,
+      },
+    });
+    return serverErrorResponse;
+  }
+
+  logger.error(LogMessage.TXMA_EVENT_SESSION_INVALID, {
```
chris-cooksley-gds, line 199: Same point here - this will be more or less redundant with `GET_SESSION_SESSION_INVALID` already logged in the adapter.

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,13 +50,184 @@ export async function lambdaHandlerConstructor(
...
+  eventData: BaseEventData;
+}): Promise<Result<void, void>> {
+  const { sessionId, issuer, ipAddress, txmaAuditEncoded } = eventData;
+  return await eventService.writeGenericEvent({
+    eventName,
+    sessionId,
+    componentId: issuer,
+    ipAddress,
+    txmaAuditEncoded,
+    sub: undefined,
```
chris-cooksley-gds, line 227: Nitpick: can we just omit the sub, journeyId, redirect_uri and fraud signal, instead of adding them in as undefined?

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
```diff
@@ -32,13 +50,184 @@ export async function lambdaHandlerConstructor(
...
+  eventService: IEventService;
+  eventData: BaseEventData;
+}): Promise<APIGatewayProxyResult> {
+  const writeEventResult = await writeEvent({
+    eventService,
+    eventName: "DCMAW_ASYNC_CRI_4XXERROR",
+    eventData,
+  });
+
+  const auditEventName = "DCMAW_ASYNC_CRI_4XXERROR";
```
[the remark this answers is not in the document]
jhumbert-dd: Agree. This is no longer applicable following the removal of writing to TxMA

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.ts
jhumbert-dd: Thanks @chris-cooksley-gds. No longer necessary following [decision to remove writing to TxMA](https://gds.slack.com/archives/C065WKXD7B6/p1742469639159639?thread_ts=1742462958.369549&cid=C065WKXD7B6)

backend-api/src/functions/common/session/getOperations/GetSessionOperation.ts
```diff
@@ -0,0 +1,25 @@
+import { AttributeValue, GetItemCommandInput } from "@aws-sdk/client-dynamodb";
+import { Result } from "../../../utils/result";
+import { SessionAttributes } from "../session";
+import {
+  ValidateSessionErrorInvalidAttributeData,
+  ValidateSessionAttributes,
+} from "../SessionRegistry";
+
+export interface GetSessionOperation {
+  getDynamoDbGetCommandInput({
```
chris-cooksley-gds, line 10: Thinking out loud: what I'm not as sure about is having `getDynamoDbGetCommandInput` here. - the only situation I can think of where this wouldn't be the same constant value, is when we're getting an active session. But that scenario is different from the others in quite a few ways (e.g. we'll need to specify the index and the sub, but not the session ID), to the point where I suspect it would be cleaner to just handle it with a different method rather than try to stretch this pattern to handle the different approach. That way we can move the command input into the adapter and the GetSessionOperations can be a bit simpler - i.e. we'll always use the session ID and main table to retrieve a session in some state, and what will vary is the attributes that session has and how they might be validated. Happy to discuss and/or defer to the group on this one.

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.ts
```diff
@@ -0,0 +1,55 @@
...
+  SessionState,
+} from "../../session";
+import { getBiometricTokenIssuedSessionAttributes } from "../../sessionAttributes/sessionAttributes";
+import { GetSessionOperation } from "../GetSessionOperation";
+import {
+  ValidateSessionErrorInvalidAttributeData,
+  ValidateSessionAttributes,
+} from "../../SessionRegistry";
+
+export class TxMAEventGetSessionOperation implements GetSessionOperation {
```
chris-cooksley-gds, line 16: Suggestion: currently this operation is named after the (only) caller that currently uses it. It might be cleaner to instead name it after what kind of session it gets - e.g. `GetBiometricTokenIssued` or `GetSessionBiometricTokenIssued`. That would make its behaviour obvious just from the name.

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  validateSession(
+    attributes: ValidateSessionAttributes,
+  ): Result<void, ValidateSessionErrorInvalidAttributeData> {
+    const { sessionState, createdAt } = attributes;
+
+    if (!sessionState || sessionState !== SessionState.BIOMETRIC_TOKEN_ISSUED) {
+      return errorResult({
+        invalidAttribute: { sessionState },
```
chris-cooksley-gds, line 43: Nitpick: very much a nice-to-have, but we could append the invalid attribute to an array in case of either of these failures. Then if the array has anything in it, we could return an errorResult with `invalidAttributes` (plural), thereby communicating everything that was wrong instead of just the first thing.

backend-api/src/functions/common/session/sessionAttributes/sessionAttributes.ts
```diff
@@ -49,6 +55,15 @@ export const getBiometricTokenIssuedSessionAttributes = (
   return successResult(sessionAttributes);
 };
 
+const isBiometricTokenIssuedSessionAttributes = (
```
chris-cooksley-gds, line 58: Praise: great way to reduce repetition

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -131,6 +143,67 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+    if (responseItem == null) {
+      return this.handleGetSessionNotFoundError({
+        errorMessage: sessionNotFound,
+      });
+    }
+
+    // Attribute type validation
+    const getSessionAttributesResult =
+      getOperation.getSessionAttributesFromDynamoDbItem(responseItem);
+    if (getSessionAttributesResult.isError) {
```
chris-cooksley-gds, line 181: So, if getting the attributes fails, we'll treat it as session not found (and down the line, the caller will get a 401, indicating that it's their fault). But if a session is corrupted/malformed in this way - e.g. one of its attributes is of the wrong type, or the session object is null, or something like that - doesn't that imply that something in _our_ system is wrong, rather than a client error? I wonder if we ought to be treating these failures as server errors (so long as we can still identify when it's a well-formed session in an incorrect state and return the right error). CC @sandymay-dd

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
```diff
@@ -42,4 +50,27 @@ export class BiometricSessionFinished implements UpdateSessionOperation {
       return getBiometricSessionFinishedSessionAttributes(item);
     }
   }
+
+  validateSession(
```
chris-cooksley-gds, line 54: How come we're adding new validation functionality to update operations as part of this ticket? I think our update operations already include session state/createdAt checks as part of their conditional expressions - and in any case I'd expect a separate ticket if we needed to change those.

backend-api/src/functions/common/session/session.ts
```diff
@@ -13,28 +13,34 @@ export type SessionAttributes =
...
 export interface BiometricSessionFinishedAttributes
   extends BaseSessionAttributes {
-  sessionState: SessionState.BIOMETRIC_SESSION_FINISHED;
   documentType: DocumentType;
   opaqueId: string;
+  sessionState: SessionState.BIOMETRIC_SESSION_FINISHED;
   biometricSessionId: string;
 }
+
+export function isOlderThan60minutes(createdAtInMilliseconds: number) {
```
chris-cooksley-gds, line 42: Currently this file is pretty much* entirely types/interfaces. Given that this function is functional, it might be better placed elsewhere.

Also, the `m` in minutes should be capitalised.

backend-api/src/functions/common/session/SessionRegistry.ts
chris-cooksley-gds, line 1: Nitpick: as we chatted about earlier - there's an argument that we could improve readability by cutting down on some of these types, without losing much of value in terms of type safety. But totally non-blocking, as it's very subjective.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -225,14 +312,65 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+      error,
+      getItemCommandInput,
+    });
+    return errorResult({
+      errorType: GetSessionError.INTERNAL_SERVER_ERROR,
+    });
+  }
+
+  private handleGetSessionNotFoundError({
+    errorMessage,
```
chris-cooksley-gds, line 343: Currently I think this error message is always the constant 'Session not found', which is information already fully communicated by the log message/message code. And I don't think the errorMessage in the result sent back to the handler is used, either. I think we can probably just remove this parameter, and the constant, and omit the errorMessage from the log/result.

backend-api/src/functions/common/session/SessionRegistry.ts
```diff
@@ -26,17 +44,83 @@ export interface SessionUpdateFailedInternalServerError {
...
   INTERNAL_SERVER_ERROR = "INTERNAL_SERVER_ERROR",
-  CONDITIONAL_CHECK_FAILURE = "CONDITIONAL_CHECK_FAILURE",
   SESSION_NOT_FOUND = "SESSION_NOT_FOUND",
+  SESSION_INVALID = "SESSION_INVALID",
 }
 
-export interface UpdateExpressionDataToLog {
-  updateExpression: string;
-  conditionExpression: string | undefined;
+export type GetSessionFailed =
```
chris-cooksley-gds, line 89: Come to think of it, I don't think we actually use any information in the caller other than whether or not it was an internal server error. I wonder if instead of a GetSessionFailed, we could literally just return a GetSessionError (and perhaps even roll SESSION_NOT_FOUND and SESSION_INVALID into one while we're at it, since they get treated in the same way).

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
jhumbert-dd: Ah yes, I think the validation method should only be on the get operation as it's not possible to get a session back if the conditional / filter expression fails. I'll remove this from the update interface

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +431,204 @@ describe("DynamoDbAdapter", () => {
...
+        });
+
+        it("Logs the failure", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode: "MOBILE_ASYNC_GET_SESSION_SESSION_NOT_FOUND",
+            errorMessage: "Session not found",
+          });
+        });
+
+        it("Returns failure with server error", () => {
```
chris-cooksley-gds, line 540: Nitpick: not a server error

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -56,6 +80,40 @@ describe("Async TxMA Event", () => {
     });
   });
 
+  describe("Config validation", () => {
+    describe.each(["SESSION_TABLE_NAME"])(
```
chris-cooksley-gds, line 84: Can we add the other two environment variables to this list?

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -121,3 +272,17 @@ const validRequest = buildRequest({
     eventName: "DCMAW_ASYNC_HYBRID_BILLING_STARTED",
   }),
 });
+
+export const mockTxmaEventSessionRegistrySuccess: SessionRegistry = {
+  updateSession: jest.fn().mockResolvedValue(
```
chris-cooksley-gds, line 277: Nitpick: I think we could also just start with an inert session registry for this - this lambda won't need a working `updateSession` implementation, and this way we won't need to maintain one. I think the same goes for some of the other handler unit tests.

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      });
+    });
+
+    describe("Given the session is valid", () => {
+      const validSessionAttributes = {
+        sessionState: SessionState.BIOMETRIC_TOKEN_ISSUED,
+        createdAt: validBiometricTokenIssuedSessionAttributes.createdAt,
+      };
+
+      it("Return and error result with the invalid attribute", () => {
```
chris-cooksley-gds, line 119: Nitpick: test name needs updating, this is a happy path.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
sandymay-dd: Hopefully I haven't got the wrong end of the stick:

If we have type/structural issues with our session record, we should return a server error.

If we have a value validation failure (e.g. session age), we return a 401 Unauthorized. 

I thought this was the pattern we'd implemented for the update operations 👀

chris-cooksley-gds: That's the right end of the stick, unless I've also got the wrong end of the stick. So, currently, it looks like we're returning a 401 unauthorized for type/structural issues (which are surfaced by `getSessionAttributesFromDynamoDbItem`. I propose that we instead return a server error in this scenario.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            errorType: GetSessionError.SESSION_NOT_FOUND,
+            session: "Session not found",
+          }),
+        );
+      });
+    });
+
+    describe("Session validation", () => {
+      describe("Given the session is in the wrong state", () => {
+        const invalidBiometricTokenIssuedSessionAttributesWrongSessionState = {
```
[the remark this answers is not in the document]
jhumbert-dd: Yeah, I planned to refactor test data. This file has been updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            );
+          result = await sessionRegistry.getSession(
+            mockSessionId,
+            getOperation,
+          );
+        });
+
+        it("Logs the failure", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode: "MOBILE_ASYNC_GET_SESSION_SESSION_NOT_FOUND",
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            errorMessage:
+              "Session could not be parsed or session validation failed",
```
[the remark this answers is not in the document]
jhumbert-dd: This now logs the invalid attribute and which will be adapted to handle multiple invalid attributes

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            session:
+              invalidBiometricTokenIssuedSessionAttributesWrongSessionState,
+          });
+        });
+
+        it("Returns failure with an invalid session error", () => {
+          expect(result).toStrictEqual(
+            errorResult({
+              errorType: GetSessionError.SESSION_NOT_FOUND,
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            .resolves(biometricTokenIssuedSessionAttributesSessionTooOldItem);
+          result = await sessionRegistry.getSession(
+            mockSessionId,
+            getOperation,
+          );
+        });
+
+        it("Logs the failure", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode: "MOBILE_ASYNC_GET_SESSION_SESSION_NOT_FOUND",
```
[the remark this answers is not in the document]
jhumbert-dd: This test case has been removed as validation branching is tested in the getOperation class

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+            session: invalidBiometricTokenIssuedSessionAttributesSessionTooOld,
+          });
+        });
+
+        it("Returns failure with an invalid session error", () => {
+          expect(result).toStrictEqual(
+            errorResult({
+              errorType: GetSessionError.SESSION_NOT_FOUND,
```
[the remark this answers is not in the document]
jhumbert-dd: [related response](https://github.com/govuk-one-login/mobile-id-check-async/pull/469/files/aa3c8a73b3513d50fe4bd5d8a6abc6392f461a9a#r2006100725)

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+              session:
+                invalidBiometricTokenIssuedSessionAttributesSessionTooOld,
+            }),
+          );
+        });
+      });
+    });
+
+    describe("Given the session was found", () => {
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+                invalidBiometricTokenIssuedSessionAttributesMissingAttributes,
+            }),
+          );
+        });
+      });
+
+      describe("Given valid session attributes were returned in response", () => {
+        const validBiometricTokenIssuedSessionAttributesItem = marshall({
+          ...validBiometricTokenIssuedSessionAttributes,
+          createdAt: 1704106860000, // 2024-01-01 11:01:00.000
```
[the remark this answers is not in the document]
jhumbert-dd: updated

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.ts
jhumbert-dd: Renamed to `GetSessionBiometricTokenIssued`

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.ts
jhumbert-dd: I've updated to return a single object containing all invalid attributes as I think it would read better in the logs. Is there a benefit to returning an array of objects which each have one invalid attribute inside? If so, it wouldn't be much effort to switch methods. cc @sandymay-dd

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
jhumbert-dd: Removed

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
jhumbert-dd: This test has been updated to return an in server error

backend-api/src/functions/common/session/updateOperations/BiometricSessionFinished/BiometricSessionFinished.ts
jhumbert-dd: Removed

backend-api/src/functions/common/session/SessionRegistry.ts
jhumbert-dd: Updated

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
jhumbert-dd: Thanks. I've removed the other two as they're not being used in this PR

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.test.ts
jhumbert-dd: Updated

backend-api/src/functions/common/session/getOperations/TxmaEvent/GetSessionBiometricTokenIssued.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  validateSession(
+    attributes: ValidateSessionAttributes,
+  ): Result<void, ValidateSessionErrorInvalidAttributesData> {
+    const { sessionState, createdAt } = attributes;
+    const invalidAttributes: ValidateSessionInvalidAttributes = {};
+
+    if (!sessionState || sessionState !== SessionState.BIOMETRIC_TOKEN_ISSUED) {
+      invalidAttributes.sessionState = sessionState;
```
sandymay-dd, line 42: non-blocking: I personally find an array cleaner to see, but it's very much a preference. 

It avoids needing to do two things to invalidAttributes (Object.entries and length) to check whether to return an error. Such a minor thing, I'm not fussed at all.

backend-api/src/functions/common/session/getOperations/TxmaEvent/TxmaEventGetSessionOperation.ts
sandymay-dd: left a comment on this one of the function itself, apologies missed this thread

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.ts
```diff
@@ -225,14 +298,59 @@ export class DynamoDbAdapter implements SessionRegistry {
...
+    getItemCommandInput,
+    sessionAttributes,
+  }: {
+    error: unknown;
+    getItemCommandInput: GetItemCommandInput;
+    sessionAttributes?: InvalidSessionAttributeTypes;
+  }): FailureWithValue<GetSessionInternalServerError> {
+    logger.error(LogMessage.GET_SESSION_UNEXPECTED_FAILURE, {
+      error,
+      getItemCommandInput,
```
jmooney-dd, line 323: issue: sometimes when we log errors, we log the operation details and other times we don't, could we make this consistent?

Was there a reason for this? IMO the operation should just be logged in the initial attempt debug log, I don't think we need it in subsequent logs but I would rather we always log it or we don't for errors. See [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/469/files#diff-31ab14319bd3be272553261c41d8f2c3dfe1e6a82fbcc17ccc6cd077e0290cd1R331-R332) (just below) for the sessionNotFound scenario where we do not log this

I've not checked other logger.errors yet in this PR but could we check that we're being consistent thought the changes

backend-api/src/functions/common/session/getOperations/TxmaEvent/GetSessionBiometricTokenIssued.ts
jhumbert-dd: Yeah, I see your point. I think the two steps in this case might be worth session cleaner logs instead of an array of objects with only one property in each. Will keep as is for now

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -46,15 +31,93 @@ export const validBaseSessionAttributes = {
...
+// Biometric token issued
+
 export const validBiometricTokenIssuedSessionAttributes = {
   ...validBaseSessionAttributes,
   sessionState: SessionState.BIOMETRIC_TOKEN_ISSUED,
   documentType: "NFC_PASSPORT",
   opaqueId: "mockOpaqueId",
 };
 
+export const validBiometricTokenIssuedSessionAttributesMobileApp = {
```
jmooney-dd, line 55: suggestion (non-blocking): 

```suggestion
export const validBiometricTokenIssuedSessionAttributes = {
```

Other than 'base' attributes, the other attribute sets have typically just followed a naming pattern based purely on the state, so I don't think it's that helpful to state `MobileApp` here

Entirely non-blocking, if you like it all good to resolve

jmooney-dd: Curious if you did think of anything I'm missing about adding MobileApp

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -46,15 +31,93 @@ export const validBaseSessionAttributes = {
...
   sessionState: SessionState.BIOMETRIC_SESSION_FINISHED,
   biometricSessionId: mockBiometricSessionId,
 };
+
+export const validBiometricSessionFinishedAttributesMobileApp = {
+  ...validBiometricSessionFinishedAttributes,
+  redirectUri: "https://www.mockRedirectUri.com",
+};
+
+// Session registry
```
jmooney-dd, line 84: suggestion (non-blocking): remove titles, it's clear by the naming and types what this is

backend-api/src/functions/testUtils/unitTestData.ts
```diff
@@ -46,15 +31,93 @@ export const validBaseSessionAttributes = {
...
+  updateSession: jest.fn(() => {
+    throw new Error("Not implemented");
+  }),
+
+  getSession: jest.fn(() => {
+    throw new Error("Not implemented");
+  }),
+};
+
+// Event service
```
jmooney-dd, line 96: suggestion (non-blocking): remove comment, it's clear by the naming and types what this is

backend-api/tests/api-tests/txmaEvent.test.ts
```diff
@@ -31,23 +37,75 @@ describe("POST /async/txmaEvent", () => {
...
 
       response = await SESSIONS_API_INSTANCE.post(
         "/async/txmaEvent",
         requestBody,
       );
+    });
 
-      console.log("RESPONSE >>>>>", response);
+    it("Returns an error and 401 status code", async () => {
+      expect(response.status).toBe(401);
```
jmooney-dd, line 58:
```suggestion
    it("Returns 401 Unauthorized response with invalid_session error", async () => {
      expect(response.status).toBe(401);
      expect(response.statusText).toBe("Unauthorized");
```

backend-api/tests/api-tests/txmaEvent.test.ts
```diff
@@ -31,23 +37,75 @@ describe("POST /async/txmaEvent", () => {
...
+      await createSessionForSub(sub);
+      sessionId = await getActiveSessionIdFromSub(sub);
+      const biometricTokenRequestBody = {
+        sessionId,
+        documentType: "NFC_PASSPORT",
+      };
+      await SESSIONS_API_INSTANCE.post(
+        "/async/biometricToken",
+        biometricTokenRequestBody,
+      );
```
jmooney-dd, line 98: suggestion (non-blocking): should this step be in an api-test helper>

backend-api/src/functions/testUtils/unitTestData.ts
jmooney-dd: Might be a few others like this in the file if you do want to remove, won't comment on anymore

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
```diff
@@ -1,26 +1,29 @@
 import { expect } from "@jest/globals";
-import "../../../tests/testUtils/matchers";
-import "dotenv/config";
 import { APIGatewayProxyResult, Context } from "aws-lambda";
+import "dotenv/config";
```
jmooney-dd, line 3: issue: should this be removed?

backend-api/src/functions/testUtils/unitTestData.ts
jhumbert-dd: As the values are different for MAM and DAD sessions, do you have a suggestion of how to distinguish them? No worries if not but as maybe appending `Mam` instead of `MobileApp` is better?

jmooney-dd: Nope that makes sense, I didn't think of that, happy to leave it as is

backend-api/tests/api-tests/txmaEvent.test.ts
jmooney-dd: We could leave api test helpers to when we write the real 200 api test

backend-api/tests/api-tests/txmaEvent.test.ts
jhumbert-dd: Updated this and the one above

backend-api/src/functions/testUtils/unitTestData.ts
jhumbert-dd: Added these in to see at a glance where new test data should live but happy to remove as the value is minimal

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
```diff
@@ -90,8 +145,101 @@ describe("Async TxMA Event", () => {
     });
   });
 
+  describe("Retrieving a session", () => {
```
jmooney-dd, line 148: Question: is this structure correct?

```
  describe("Retrieving a session", () => {
    describe("Given there is an error retrieving the session", () => {
    describe("Given failure is due to client error", () => {
    });
  });
```

I don't think they should be at the same level as `Given failure is due to client error` is still still part of `Given there is an error retrieving the session`

jmooney-dd: Follow-up question re this, the error naming confuses me a bit, because one is for a client error and one is for unexpected internal server error but we often consider the latter to be the former.

I can see here where you have enum

export enum GetSessionError {
  INTERNAL_SERVER_ERROR = "INTERNAL_SERVER_ERROR",
  CLIENT_ERROR = "CLIENT_ERROR",
}

If this makes sense to you then happy to not question this anymore for this commit

backend-api/src/functions/testUtils/unitTestData.ts
jhumbert-dd: Removed

backend-api/src/functions/testUtils/unitTestData.ts
jhumbert-dd: Removed

backend-api/src/functions/asyncTxmaEvent/asyncTxmaEventHandler.test.ts
jhumbert-dd: Updated

backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.test.ts
jhumbert-dd: Updated

backend-api/tests/api-tests/txmaEvent.test.ts
jhumbert-dd: updated

backend-api/src/functions/common/session/SessionRegistry.ts
sandymay-dd: We agree that there's type proliferation. A not for now one, given you said it's non-blocking.

backend-api/src/functions/adapters/aws/dynamo/dynamoDbAdapter.test.ts
```diff
@@ -293,4 +309,234 @@ describe("DynamoDbAdapter", () => {
...
+              "Session could not be parsed or session validation failed",
+            session:
+              invalidBiometricTokenIssuedSessionAttributesMissingAttributes,
+          });
+        });
+
+        it("Returns failure with server error", () => {
+          expect(result).toEqual(
+            errorResult({
+              errorType: GetSessionError.SESSION_NOT_FOUND,
```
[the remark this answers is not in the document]
jhumbert-dd: Updated

sandymay-dd (approved):
Approved the last few commits having taken over James' review.

PR #470, opened 2025-03-10
DCMAW-11113: Updates finishBiometricSession to write ASYNC_APP_END TxMA event

jmooney-dd:
### What changed

- Writes `DCMAW_ASYNC_APP_END` in `async/finishBiometricSession` when request received with valid session information
- If event fails to write, return `500 Internal server error` response
- Adds `DCMAW_ASYNC_APP_END` event to dequeue lambda `getEvents` list so it can be picked up and API tested (see below re the API test)

### What didn't change

- I have not added an API test for this change as we are covering the happy path API under this ticket. This requirement has been split into a different ticket as we want to have a dev discussion re the shape of this happy with API test.

### Approach

You'll see I haven't used a helper function in this lambda for this event. My thinking was the two main things this lambda does is write the vendor processing message, and write the app end event, therefore I don't mind having these two things very clearly laid out in the lambda without handler functions. I am happy to discuss this and change if you feel any helper functions would provide value here.

### Why did it change

Auditing purposes

### Automated

Deployed and ran API tests - all passing

### Manual

As there is no happy path API test for this endpoint, the API tests will not currently trigger this event so I did it manually.

Screenshot 2025-03-10 at 11 11 39

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
+++ b/backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
@@ -555,6 +555,57 @@ describe("Async Finish Biometric Session", () => {
       );
     });
 
+    describe("Given sending DCMAW_ASYNC_APP_END event fails", () => {
+      beforeEach(async () => {
+        dependencies = {
+          ...dependencies,
+          getEventService: () => mockFailingEventService,
+        };
+
+        result = await lambdaHandlerConstructor(
+          dependencies,
+          validRequest,
+          context,
+        );
+      });
+
+      it("Logs the DCMAW_ASYNC_APP_END event failure", () => {
+        expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+          messageCode: "MOBILE_ASYNC_ERROR_WRITING_AUDIT_EVENT",
+          data: {
+            auditEventName: "DCMAW_ASYNC_APP_END",
+          },
+        });
+      });
+
+      it("Returns 500 Internal server error", () => {
+        expect(result).toStrictEqual({
+          statusCode: 500,
+          body: JSON.stringify({
+            error: "server_error",
+            error_description: "Internal Server Error",
+          }),
+          headers: expectedSecurityHeaders,
+        });
+      });
+    });
+
+    it("Writes DCMAW_ASYNC_APP_END event to TxMA", () => {
+      expect(mockWriteGenericEventSuccess).toBeCalledWith({
+        eventName: "DCMAW_ASYNC_APP_END",
+        sub: "mockSubjectIdentifier",
+        sessionId: "mockSessionId",
+        govukSigninJourneyId: "mockGovukSigninJourneyId",
+        componentId: "mockIssuer",
+        transactionId: "f32432a9-0965-4da9-8a2c-a98a79349d4a",
+        redirect_uri: undefined,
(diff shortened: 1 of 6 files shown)


backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.ts
```diff
@@ -83,22 +83,45 @@ export async function lambdaHandlerConstructor(
...
+    ipAddress,
+    txmaAuditEncoded,
+  });
+  if (writeAppEndEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: {
+        auditEventName: "DCMAW_ASYNC_APP_END",
+      },
+    });
+    return serverErrorResponse;
```
jmooney-dd, line 122: thought: I think we could make a nice helper function in our lambdaResponses file for these internal error responses where we log an audit event has failed as we do it several places over the codebase. 

If people like this idea I can do it as a tiny follow-up PR, I don't want to add it to this commit

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
```diff
@@ -555,6 +555,57 @@ describe("Async Finish Biometric Session", () => {
...
+
+    it("Writes DCMAW_ASYNC_APP_END event to TxMA", () => {
+      expect(mockWriteGenericEventSuccess).toBeCalledWith({
+        eventName: "DCMAW_ASYNC_APP_END",
+        sub: "mockSubjectIdentifier",
+        sessionId: "mockSessionId",
+        govukSigninJourneyId: "mockGovukSigninJourneyId",
+        componentId: "mockIssuer",
+        transactionId: "f32432a9-0965-4da9-8a2c-a98a79349d4a",
+        redirect_uri: undefined,
```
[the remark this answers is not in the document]
jmooney-dd: It's passed but the request in this test doesn't include it therefore undefined, I'll add it into the request!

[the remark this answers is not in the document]
jmooney-dd: Updated

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
```diff
@@ -555,6 +558,57 @@ describe("Async Finish Biometric Session", () => {
...
+          headers: expectedSecurityHeaders,
+        });
+      });
+    });
+
+    it("Writes DCMAW_ASYNC_APP_END event to TxMA", () => {
+      expect(mockWriteGenericEventSuccess).toBeCalledWith({
+        eventName: "DCMAW_ASYNC_APP_END",
+        sub: "mockSubjectIdentifier",
+        sessionId: "mockSessionId",
```
jhumbert-dd, line 600: suggestion: there's a common `mockSessionId` that could be used here.

jhumbert-dd (approved):

backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
jmooney-dd: Updated

jhumbert-dd (approved):

PR #474, opened 2025-03-11
DCMAW-11115: Updates finishBiometricSession to return 200 OK response

jmooney-dd:
### What changed

- Updates `finishBiometricSession` to return `200 OK` response
- Updates `okResponse` function to cater for when the `response` `body` is empty
- Adds API test for `200` response, including assertion on the `DCMAW_ASYNC_APP_END` TxMA event
- API test helpers
  - Removes duplication in API test helpers that got a `sessionId`
  - Regarding the `sessionId` methods, the main one used was doing too much as it was creating a session with a sub and returning a sessionId, I have made it so a test will use a bespoke function for each step so it is clearer. 
  - Updated biometricToken api tests to use new helper functions where appropriate
- Minor structural updates to finishBiometricSession API tests (beforeAll() and then only assertions in the it which is the pattern we are currently using)

### Why did it change

Ongoing work to develop the `finishBiometricSession` lambda.

## When reviewing

Sandy wanted us as a team to discuss the API test approach regarding the API helper functions. 

He brought particular attention to the `getValidSessionId` function as something we might want to change. We have not been able to have this conversation as a dev team so I would ask that you specifically think about the API helper functions when reviewing. Do they make sense to you? Is the naming ok? Any changes/additions you would make? Are they used how you'd expect?

### Automated

Deployed stack and ran API tests - all passing

## Checklists

### DIFF
--- a/backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
+++ b/backend-api/src/functions/asyncFinishBiometricSession/asyncFinishBiometricSessionHandler.test.ts
@@ -539,7 +539,7 @@ describe("Async Finish Biometric Session", () => {
     });
   });
 
-  describe("Given a valid request is made", () => {
+  describe("Given a request containing a valid sessionId is made", () => {
     beforeEach(async () => {
       result = await lambdaHandlerConstructor(
         dependencies,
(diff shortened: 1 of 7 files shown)


tmulberry-gds (approved):
LGTM

jhumbert-dd (approved):

PR #477, opened 2025-03-12
[DCMAW-11501] Create Subscription Filters for forward logs to Splunk

mthkbarber:
​DCMAW-11501

### What changed
Set up Subscription Filters to forward logs to CSLS in Build and above environments.
Subscription Filters will deploy in Dev and Build for proxy resources.
Add a test to check that all Cloudwatch log groups have an associated filter.


backend-api/infra/api/proxy.yaml
```diff
@@ -42,6 +42,14 @@ Resources:
...
     Condition: ProxyApiDeployment
 
+  ProxyApiSubscriptionFilter:
+    Type: AWS::Logs::SubscriptionFilter    
+    Properties:
+      DestinationArn: arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2
+      FilterPattern: ""
+      LogGroupName: !Ref ProxyApiAccessLogs
+    Condition: ProxyApiDeployment
+
```
mthkbarber, line 52: These deploy in Dev and Build as I didn't want to overcomplicate the conditionals as the Proxy Api and Lambda only deploy in Dev and Build

backend-api/infra/api/private.yaml
```diff
@@ -54,4 +54,12 @@ Resources:
...
\ No newline at end of file
+      RetentionInDays: 30
+
+  AsyncCredentialPrivateApiSubscriptionFilter:    
+    Type: AWS::Logs::SubscriptionFilter
+    Properties:
+      DestinationArn: arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2
+      FilterPattern: ""
+      LogGroupName: !Ref AsyncCredentialPrivateApiAccessLogs
+    Condition: isNotDev
```
sandymay-dd, line 65: nit: is it please possible to create a new condition for this work? A bit like: https://github.com/govuk-one-login/mobile-id-check-async/blob/3fca1d1dde9a7653e7787398aaa10c45a154186e/backend-api/template.yaml#L233

egressLogsToSplunk or deployCSLSSubscriptionFilter perhaps

backend-api/infra/api/private.yaml
```diff
@@ -54,4 +54,12 @@ Resources:
...
     Properties:
       LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-private-api-access-logs
-      RetentionInDays: 30
\ No newline at end of file
+      RetentionInDays: 30
+
+  AsyncCredentialPrivateApiSubscriptionFilter:    
+    Type: AWS::Logs::SubscriptionFilter
+    Properties:
+      DestinationArn: arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2
```
sandymay-dd, line 62: nit[not-blocking: it might be nice for this to be a Mapping. e.g. https://github.com/govuk-one-login/mobile-id-check-async/blob/3fca1d1dde9a7653e7787398aaa10c45a154186e/backend-api/template.yaml#L215

backend-api/infra/api/private.yaml
```diff
@@ -54,4 +54,12 @@ Resources:
     Type: AWS::Logs::LogGroup
     Properties:
       LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-private-api-access-logs
-      RetentionInDays: 30
\ No newline at end of file
+      RetentionInDays: 30
+
+  AsyncCredentialPrivateApiSubscriptionFilter:
```
sandymay-dd, line 59: issue: the Async**Credential** in the resource name is a legacy mistake. Are we please ok to remove 'Credential' from the logical id here, and in the LogGroup logical id in the resource above?

I don't know how widely impacting this is

backend-api/infra/api/private.yaml
mthkbarber: How do you feel about `ProdLikeEnvironment` like V1? As I feel like we will probably end up needing to use a condition for Build, Staging, Int and Prod elsewhere?

backend-api/tests/infra-tests/application.test.ts
```diff
@@ -684,6 +684,16 @@ describe("Backend application infrastructure", () => {
       });
     });
 
+    test("All log groups have a CSLS subscription filter", () => {
+      const log_groups = template.findResources("AWS::Logs::LogGroup");
+      const logs_list = Object.keys(log_groups);
+      logs_list.forEach((log_name) => {
+        template.hasResourceProperties("AWS::Logs::SubscriptionFilter", {
```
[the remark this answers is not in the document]
mthkbarber: Will this work if we change it to mappings?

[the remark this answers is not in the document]
sandymay-dd: The test will need updating, but yes it will :)

backend-api/infra/api/private.yaml
sandymay-dd: I misunderstood the principle behind some of our Condition naming. Here it's pretty clear why we're deploying a resource, so the isNotDev is 👍

backend-api/infra/api/private.yaml
```diff
@@ -50,8 +50,16 @@ Resources:
...
       LogGroupName: !Sub /aws/apigateway/${AWS::StackName}-private-api-access-logs
-      RetentionInDays: 30
\ No newline at end of file
+      RetentionInDays: 30
+
+  AsyncPrivateApiSubscriptionFilter:    
+    Type: AWS::Logs::SubscriptionFilter
+    Properties:
+      DestinationArn: arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2
+      FilterPattern: ""
```
kikidawson-gds, line 63: Does this mean that all logs get sent to the log group?

mthkbarber: Yeah they then get processed into Splunk.

backend-api/infra/api/private.yaml
```diff
@@ -50,8 +50,17 @@ Resources:
           Parameters:
             Location: ./openApiSpecs/async-private-spec.yaml
 
-  AsyncCredentialPrivateApiAccessLogs:
+  AsyncPrivateApiAccessLogs:
```
mthkbarber, line 53: This will block the pipeline @sandymay-dd as CFN will try to create a new resource but it will need a new LogGroupName, am I okay to update the `LogGroupName` to prevent this happening?

mthkbarber: My suggestion is to change it, and then change it back?

backend-api/infra/api/proxy.yaml
mthkbarber: Had to change this as it blocks dev stacks due to IAM permissions.

backend-api/infra/functions/proxy.yaml
```diff
@@ -36,6 +36,14 @@ Resources:
...
     Condition: ProxyApiDeployment
 
+  ProxyLambdaSubscriptionFilter:    
+    Type: AWS::Logs::SubscriptionFilter
+    Properties:
+      DestinationArn: arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2
+      FilterPattern: ""
+      LogGroupName: !Ref ProxyLambdaLogGroup
+    Condition: ProxyApiDeployment
+
```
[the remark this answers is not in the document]
mthkbarber: Had to change this as it blocks dev stacks due to IAM permissions.

backend-api/infra/api/private.yaml
kikidawson-gds: this is where we need terraform to allow us to delete from creation

kikidawson-gds: This is still a 2 step process but means we can have the correct log group name from the start.
We could deploy with a change to the old log group name first and get that change merged first. Then we can have this resource with the correct log group name in this PR

backend-api/infra/functions/proxy.yaml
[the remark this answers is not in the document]
kikidawson-gds: Sounds good to me

backend-api/infra/functions/jwks.yaml
```diff
@@ -93,6 +93,15 @@ Resources:
       RetentionInDays: 30
       LogGroupName: !Sub /aws/lambda/${AWS::StackName}-jwks
 
+  JsonWebKeysFunctionSubscriptionFilter:
```
kikidawson-gds, line 96: Discussion (no-blocking): Do you think we should create a logical ID naming convention. Possibly a bit late now but better now than when we're live. Some of the subscription filter names have function, some have, lambda, some don't have either.

mthkbarber: I noticed the same thing!
I have just used the same logical ID structure as the associated CloudWatch Log Group for now.

kikidawson-gds (approved):

jmooney-dd:
Checked with @mthkbarber happy to close

PR #482, opened 2025-03-14
DCMAW-11501 Update log group name so it can revert back in the other PR

mthkbarber:
​DCMAW-11501

### What changed
Change log group name so we can update it in [pull request here](https://github.com/govuk-one-login/mobile-id-check-async/pull/477)

### Why did it change
We are going to change the logical ID and it will create a new resource log group before deleting the old one, so we need to change the log group name otherwise the stack update will fail.

### DIFF
--- a/backend-api/infra/api/private.yaml
+++ b/backend-api/infra/api/private.yaml
@@ -28,7 +28,7 @@ Resources:
       StageName: !Ref Environment
       OpenApiVersion: 3.0.1
       AccessLogSetting:
-        DestinationArn: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:${AsyncCredentialPrivateApiAccessLogs}-v2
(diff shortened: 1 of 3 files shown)


backend-api/infra/api/private.yaml
```diff
@@ -28,7 +28,7 @@ Resources:
       StageName: !Ref Environment
       OpenApiVersion: 3.0.1
       AccessLogSetting:
-        DestinationArn: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:${AsyncCredentialPrivateApiAccessLogs}
+        DestinationArn: !Sub arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:${AsyncCredentialPrivateApiAccessLogs}-v2
```
kikidawson-gds, line 31: do we need this `-v2`?

mthkbarber: No. Thanks!

kikidawson-gds (approved):

jmooney-dd (approved):

PR #569, opened 2025-04-08
DCMAW-12240: pre refactor

jmooney-dd:
No need to review, raised for a discussion on Wednesday morning


backend-api/src/functions/common/session/getOperations/GetSessionIssueBiometricCredential/GetSessionIssueBiometricCredential.ts
```diff
@@ -0,0 +1,49 @@
...
+import { isOlderThan60Minutes } from "../../../../utils/utils";
+
+export class GetSessionIssueBiometricCredential implements GetSessionOperation {
+  getSessionAttributesFromDynamoDbItem(
+    item: Record<string, AttributeValue>,
+  ): Result<
+    BiometricSessionFinishedAttributes,
+    ValidateSessionErrorInvalidAttributeTypeData
+  > {
+    return myFuncTempName(item);
```
jmooney-dd, line 24: I can't use `getBiometricSessionFinishedSessionAttributes` so I had to make a new func which for now is called myFuncTempName

I can't use `getBiometricSessionFinishedSessionAttributes` because:
- getSession checks if Item is null before passing whereas updateSession does not check if item is null before passing
- Different return types, getSession returns attributes to log, whereas updateSession does not

I could just have this separate func with a better name but seems duplicative

PR #586, opened 2025-04-15
DCMAW-12243: Issuing VC | Get Biometric Session from ReadID

tmulberry-gds:
## Jira Ticket

DCMAW-12243

## Review guidance
<details>
<summary>Review checklist</summary>

### Functional Review
- **Functionality**: Does it meet the acceptance criteria on the ticket and work as expected?
- **Requirements**: Does the code meet functional and non-functional requirements including compliance with programme standards and security gates?

### Security & Compliance
- **Personally Identifiable Information**: Is it possible for PII to be logged?
- **Security Considerations**: Are there any security implications that need to be addressed?

### Quality Assurance
- **Testing**: Is the code well-tested with sufficient coverage to provide confidence in correctness?
- **Edge Cases**: Have edge cases been considered and handled appropriately?

### Code Quality
- **Readability**: Is the code easy to understand for all team members, with clear naming and appropriate documentation?
- **Maintainability**: Is the code easy to change, reuse, and extend?
- **Code Style**: Does it follow our coding conventions and best practices?
- **Code Quality**: Is the code maintainable and following best practices? See [Values, Principles & Practices](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/4955865126/ID+Check+Web+-+Values+Principles+and+Practices)

### Observability & Operations
- **Observability**: Are there appropriate logs/metrics that would help debug and monitor the service?
- **Performance**: Are there any performance considerations or potential bottlenecks?

### Documentation
- **Documentation**: Is the code well documented? Is there any existing documentation that needs updating?
- **Comments**: Are complex sections of code adequately commented if the intent is not clear?

### Review PR:
- **Title**: Contains ticket number and clear summary of change
- **Description**: Has clear description of change
</details>


backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +417,153 @@ describe("Async Issue Biometric Credential", () => {
...
+      ).rejects.toThrow(/Biometric session not ready: PROCESSING/);
+
+      expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
+        messageCode: "MOBILE_ASYNC_BIOMETRIC_SESSION_NOT_READY",
+        data: {
+          finish: "PROCESSING",
+        },
+      });
+    });
+  });
```
jmooney-dd, line 437: suggestion: see other tests in this file that use try catch in the before each, this way we can still separate the assertions, one it block for log, one for the error

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +90,88 @@ export async function lambdaHandlerConstructor(
   if (viewerKeyResult.isError) {
     throw new RetainMessageOnQueue("Failed to retrieve biometric viewer key");
   }
+  const submitterKey = viewerKeyResult.value;
+
+  const sessionResult = await dependencies.getBiometricSession(
```
jmooney-dd, line 95: issue: could this const be named something idfferently so it's not confused with any session result when getting OUR session

```suggestion
  const biometricSessionResult = await dependencies.getBiometricSession(
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +90,88 @@ export async function lambdaHandlerConstructor(
...
+    config.READID_BASE_URL,
+    sessionId,
+    submitterKey,
+  );
+
+  if (sessionResult.isError) {
+    const eventService = dependencies.getEventService(config.TXMA_SQS);
+
+    // Check if the error was retryable based on HTTP status code
+    if (isRetryableError()) {
```
jmooney-dd, line 105: question: does this work and is this best practice?

When I read this, I wonder what it is evaluating. When clicking through I can see it's the let lastError variable in another file. Was there a reason for this approach or could we pass this func something to evaluate?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -38,19 +45,28 @@ export async function lambdaHandlerConstructor(
...
   const sessionRegistry = dependencies.getSessionRegistry(
     config.SESSION_TABLE_NAME,
   );
+
   const eventService = dependencies.getEventService(config.TXMA_SQS);
 
   const getSessionResult = await sessionRegistry.getSession(
     sessionId,
-    new GetSessionIssueBiometricCredential(),
+    new GetSessionBiometricTokenIssued(),
```
jmooney-dd, line 61: issue: this is the name of a different lambda

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,136 @@
...
+  baseUrl: string,
+  sessionId: string,
+  submitterKey: string,
+  sendHttpRequest?: ISendHttpRequest,
+) => Promise<Result<BiometricSession, void>>;
+
+export const getBiometricSession: GetBiometricSession = async (
+  baseUrl: string,
+  sessionId: string,
+  submitterKey: string,
```
chris-cooksley-gds, line 44:
```suggestion
  viewerKey: string,
```

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,136 @@
...
+export type GetBiometricSession = (
+  baseUrl: string,
+  sessionId: string,
+  submitterKey: string,
+  sendHttpRequest?: ISendHttpRequest,
+) => Promise<Result<BiometricSession, void>>;
+
+export const getBiometricSession: GetBiometricSession = async (
+  baseUrl: string,
+  sessionId: string,
```
chris-cooksley-gds, line 43:
```suggestion
  biometricSessionId: string,
```

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,136 @@
...
+    method: "GET" as const,
+    headers: {
+      "X-Innovalor-Authorization": submitterKey,
+      Accept: "application/json;odata.metadata=minimal",
+      "Content-Type": "application/json;odata.metadata=minimal",
+      "OData-MaxVersion": "4.0",
+      "OData-Version": "4.0",
+    },
+  };
+  const httpRequestLogData = {
```
chris-cooksley-gds, line 58: Do we need to log any of this data at all? The only things that are dynamic are:
* the biometric session ID - which I suppose we should already be logging elsewhere (but if we're not I'd be fine with logging it here)
* the viewer key, which gets redacted anyway

Everything else (url, header values etc) should always be the same - and therefore won't give us any information we don't already have.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,136 @@
...
+    ...httpRequest,
+    headers: {
+      ...httpRequest.headers,
+      "X-Innovalor-Authorization": "Secret value and cannot be logged",
+    },
+  };
+  const retryConfig = {
```
chris-cooksley-gds, line 65: This doesn't look like it matches the `RetryConfig` type expected by `sendHttpRequest` - is typescript ok with this?

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,136 @@
...
+      httpRequest: httpRequestLogData,
+      sessionId,
+    },
+  });
+
+  const getBiometricSessionResult: Result<SuccessfulHttpResponse, HttpError> =
+    await sendHttpRequest(httpRequest, retryConfig);
+
+  if (getBiometricSessionResult.isError) {
+    lastError = getBiometricSessionResult.value;
```
chris-cooksley-gds, line 86: Rather than using this `lastError` variable to share state with the handler, we could:
* return an error object instead of an empty failure from this function, which would indicate whether the failure was retryable or not
* just throw the RetainMessageOnQueue exception here if the failure is retryable

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +90,88 @@ export async function lambdaHandlerConstructor(
...
+
+    return;
+  }
+
+  const session = sessionResult.value;
+
+  if (session.finish !== "DONE") {
+    logger.info(LogMessage.BIOMETRIC_SESSION_NOT_READY, {
+      data: { finish: session.finish },
+    });
```
chris-cooksley-gds, line 168: I _think_ we should also log completion here, per the ticket. The rationale being that this isn't an error scenario, so we wouldn't want it to drive canary rollbacks etc.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +417,153 @@ describe("Async Issue Biometric Credential", () => {
     });
   });
 
+  describe("When biometric session is not ready", () => {
+    beforeEach(() => {
+      dependencies.getBiometricSession = mockGetBiometricSessionNotReady;
```
chris-cooksley-gds, line 422: Based on our recent Dev Time discussion around unit testing and mocking - do we need to mock this, and have separate unit tests directly testing `getBiometricSession`? I think the pattern we agreed on for cases like this (where the dependency isn't shared, or an adapter), was to test through the handler tests - splitting into a separate file if needed for readability.

This comment is non-blocking though as it's a new pattern for our test structuring and this would be the first time using it.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -38,19 +45,28 @@ export async function lambdaHandlerConstructor(
...
+
   const eventService = dependencies.getEventService(config.TXMA_SQS);
 
   const getSessionResult = await sessionRegistry.getSession(
     sessionId,
-    new GetSessionIssueBiometricCredential(),
+    new GetSessionBiometricTokenIssued(),
   );
+
+  if (getSessionResult.isError) {
```
chris-cooksley-gds, line 64: Why has this been added? I think our session error handling (including logging) should already be complete.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: Seconding this point - see my comment in `getBiometricSession` about other ways to go about this.

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -311,6 +311,45 @@ export class LogMessage implements LogAttributes {
     "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_INVALID_CONFIG",
     "One or more required environment variables were missing or invalid.",
   );
+
+  static readonly BIOMETRIC_SESSION_GET_FROM_READID_FAILURE = new LogMessage(
+    "MOBILE_ASYNC_BIOMETRIC_SESSION_GET_FROM_READID_FAILURE",
```
chris-cooksley-gds, line 316: Can we follow the same pattern as our other lambda-specific logs? (i.e. '**ISSUE_BIOMETRIC_CREDENTIAL_**EXAMPLE_LOG')

backend-api/src/functions/asyncIssueBiometricCredential/RetainMessageOnQueue.ts
```diff
@@ -1,8 +1,16 @@
-type RetainMessageOnQueueMessages =
+type StaticRetainMessages =
   | "Invalid config"
   | "Failed to retrieve biometric viewer key"
   | "Unexpected failure retrieving session from database";
 
+type RetryableErrorMessage =
+  | `Retryable error (${number}) retrieving biometric session`
+  | `Biometric session not ready: ${string}`;
```
chris-cooksley-gds, line 8: Nitpick: I know technically it's used for a message in an `Error` subclass (because of how we've implemented retries) - but it doesn't feel right calling this second one an _error message_ as the session not being ready isn't really an error, it's just a normal thing that can happen sometimes.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Changed to biometricSessionId

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Yeah, I was planning to use it at least Yeah I can remove it, was mostly just going to be for my own confidence when running it. Removed

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Added completion log

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Removed

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: See comment above

backend-api/src/functions/asyncIssueBiometricCredential/RetainMessageOnQueue.ts
tmulberry-gds: I've split it up further into a 'NonErrorMessage' type for clarity, what are your thoughts?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Good point. I didn't realise we were going to adapt this session so soon, but I'm happy we are.

The tests in asyncIssueBiometricCredentialHandler.test.ts that check "When biometric session is not ready" and "When biometric session retrieval fails" are primarily testing the handler's behaviour when it receives different responses from the getBiometricSession function, rather than testing the getBiometricSession functionality itself.

I'm struggling to remember what the exact outcome of the dev time was, as in would we test lambda specific helpers inside the lambda itself, or extract it out. Happy to discuss this though

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Updated

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Updated naming

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -7,17 +7,33 @@ import { lambdaHandlerConstructor } from "./asyncIssueBiometricCredentialHandler
...
 import { GetSessionError } from "../common/session/SessionRegistry/types";
+import { BiometricSession } from "./getBiometricSession/getBiometricSession";
+
+jest.mock("./getBiometricSession/getBiometricSession", () => {
+  return {
+    ...jest.requireActual("./getBiometricSession/getBiometricSession"),
+    isRetryableError: jest.fn(),
+    getLastError: jest.fn(),
+  };
+});
```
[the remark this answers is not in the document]
tmulberry-gds: Yeah this is just a mistake on my part lol. Removed

backend-api/src/functions/asyncIssueBiometricCredential/RetainMessageOnQueue.ts
```diff
@@ -1,8 +1,16 @@
-type RetainMessageOnQueueMessages =
+type StaticRetainMessages =
```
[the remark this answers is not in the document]
chris-cooksley-gds: I'd vote to keep them all in a single type for simplicity - but that's pretty weakly held

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds: So the idea would be something like:

* update the handler to add `sendHttpRequest` into its dependencies, and pass this into `getBiometricSession.ts` so that we can provide a mock implementation of it in the handler unit tests
* remove the `getBiometricSession.test.ts` suite
* add tests to the handler test suite covering the same scenarios in the deleted suite, but calling the handler instead of calling `getBiometricSession` directly (e.g. we'd put the relevant mockSendHttpRequest into the handler dependencies, and then call the handler with a valid payload and assert on what was thrown/logged)
* if desired, split out those tests into their own file for readability - but crucially they're still just invoking the handler, and `getBiometricSession.ts` never gets called by any test.

But again this is non-blocking as we've only just decided on this pattern - given proximity to the end of sprint, it might be best to keep this as-is for now and look to follow this pattern next time.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
chris-cooksley-gds: Unresolving this as it's still being logged - is it possible that a change hasn't pushed to github yet?

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,171 @@
...
+          error,
+          httpRequest: httpRequestLogData,
+          biometricSessionId,
+        },
+      },
+    );
+
+    return errorResult({
+      message: "Failed to parse response JSON",
+      isRetryable: true,
```
chris-cooksley-gds, line 140: If ReadID sends a malformed response (including the malformed JSON here, or the empty response body above), I'm not sure that's something we should retry. But Tagging @sandymay-dd as I'm not confident on that point.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,91 @@ export async function lambdaHandlerConstructor(
   if (viewerKeyResult.isError) {
     throw new RetainMessageOnQueue("Failed to retrieve biometric viewer key");
   }
+  const submitterKey = viewerKeyResult.value;
```
chris-cooksley-gds, line 82:
```suggestion
  const viewerKey = viewerKeyResult.value;
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,91 @@ export async function lambdaHandlerConstructor(
...
+    submitterKey,
+  );
+
+  if (biometricSessionResult.isError) {
+    const eventService = dependencies.getEventService(config.TXMA_SQS);
+    const error: BiometricSessionError = biometricSessionResult.value;
+
+    // Check if the error was retryable based on error info
+    if (error.isRetryable) {
+      logger.error(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_RETRYABLE_ERROR);
```
chris-cooksley-gds, line 96: This log feels redundant, since we've already logged the error (and the value of `isRetryable`) in `getBiometricSession`

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,91 @@ export async function lambdaHandlerConstructor(
...
+      throw new RetainMessageOnQueue(
+        `Retryable error (${error.statusCode || "unknown"}) retrieving biometric session`,
```
chris-cooksley-gds, line 98:
```suggestion
        `Retryable error (status code: ${error.statusCode || "N/A"}) retrieving biometric session`,
```

^ I think this gives a clearer result when there's no status code (i.e. we didn't get an HTTP response due to a network error or similar)

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
sandymay-dd: Yep malformed response would not be retryable.

backend-api/src/functions/asyncIssueBiometricCredential/RetainMessageOnQueue.ts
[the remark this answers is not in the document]
tmulberry-gds: Okay, changed them to a single type

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Changed to viewerKey

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Changed

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Removed the log

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Yeah change wasn't pushed. got distracted by a failing test

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+      "OData-MaxVersion": "4.0",
+      "OData-Version": "4.0",
+    },
+  };
+
+  const retryConfig: RetryConfig = {
+    retryableStatusCodes: [
+      429, 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, 511,
+    ],
+    maxAttempts: 3,
```
chris-cooksley-gds, line 52: Nitpick: we can omit `maxAttempts` here as the default value is already `3`.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: Needs pushing

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: Needs pushing

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: Needs pushing

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+  const getBiometricSessionResult: Result<SuccessfulHttpResponse, HttpError> =
+    await sendHttpRequest(httpRequest, retryConfig);
+
+  if (getBiometricSessionResult.isError) {
+    const error = getBiometricSessionResult.value;
+    const statusCode = error.statusCode ? Number(error.statusCode) : undefined;
+
+    const isRetryable =
+      !statusCode ||
+      statusCode === 429 ||
```
chris-cooksley-gds, line 71: Suggestion: we could extract the `retryableStatusCodes` above into a const array, and then refer to that here to avoid duplication (e.g. `isRetryable = !statusCode || retryableStatusCodes.includes(statusCode)`)

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,91 @@ export async function lambdaHandlerConstructor(
   if (viewerKeyResult.isError) {
     throw new RetainMessageOnQueue("Failed to retrieve biometric viewer key");
   }
+  const submitterKey = viewerKeyResult.value;
+
+  const biometricSessionResult = await dependencies.getBiometricSession(
+    config.READID_BASE_URL,
+    sessionId,
```
chris-cooksley-gds, line 86: issue: this should be the `biometricSessionId` - you'll need to pull that out of the `sessionAttributes` above.

Once you've pulled it out - you might as well add it to the logger (`logger.appendKeys({biometricSessionId})`), and then you don't need to add it to the log data every time.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+  SuccessfulHttpResponse,
+} from "../../adapters/http/sendHttpRequest";
+import { Result, successResult, errorResult } from "../../utils/result";
+
+export interface BiometricSession {
+  id: string;
+  finish: string;
+}
+
+export interface BiometricSessionError {
```
chris-cooksley-gds, line 17:
```suggestion
export interface GetBiometricSessionError {
```

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+import {
+  HttpError,
+  ISendHttpRequest,
+  RetryConfig,
+  sendHttpRequest as sendHttpRequestDefault,
+  SuccessfulHttpResponse,
+} from "../../adapters/http/sendHttpRequest";
+import { Result, successResult, errorResult } from "../../utils/result";
+
+export interface BiometricSession {
```
chris-cooksley-gds, line 12: Thoughts:
* I don't think we do anything in this function to prove that there's an `id` field - nor do we depend on such a field - so it can probably be removed.
* A bit pedantic - and very much non-blocking - but I'd prefer that this not be called `BiometricSession`, mainly because (a) we haven't yet proved that it's a valid BiometricSession (that will happen in the package), and (b) we use that type name in the package to refer to a full-fledged BiometricSession with all the properties. Maybe we could go with something like `VendorResponse`?

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+  id: string;
+  finish: string;
+}
+
+export interface BiometricSessionError {
+  statusCode?: number;
+  message: string;
```
chris-cooksley-gds, line 19: I don't think this field ever gets used - in which case it can be removed.

I'd also argue that we can remove `statusCode`. Since we've already logged it in this function, the only purpose of passing it back it to have it appear in the error message, which I don't think we need (I think it's enough to know we got a retryable failure, we can check the logs easily if we need more detail than that)

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +420,119 @@ describe("Async Issue Biometric Credential", () => {
...
+        it("Logs IPV Core message error", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode:
+              "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR",
+          });
+        });
+
+        it("Still sends event to TxMA", () => {
+          expect(mockWriteGenericEventSuccessResult).toHaveBeenCalled();
+        });
```
jmooney-dd, line 532: question: is this not tested [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/586/files#diff-0943d2a2aba10106cb60b120be197df37495fcee2b70caa718c8485079ec7ca5R504)?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +420,119 @@ describe("Async Issue Biometric Credential", () => {
...
+
+        it("Logs IPV Core message error", () => {
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode:
+              "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR",
+          });
+        });
+
+        it("Still sends event to TxMA", () => {
```
jmooney-dd, line 530: suggestion: should this test read more like [this one](https://github.com/govuk-one-login/mobile-id-check-async/pull/586/files#diff-0943d2a2aba10106cb60b120be197df37495fcee2b70caa718c8485079ec7ca5R504)?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,91 @@ export async function lambdaHandlerConstructor(
...
+      logger.error(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_RETRYABLE_ERROR);
+      throw new RetainMessageOnQueue(
+        `Retryable error (${error.statusCode || "unknown"}) retrieving biometric session`,
+      );
+    }
+
+    // Non-retryable error - send error to IPV Core
+    logger.error(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_NON_RETRYABLE_ERROR, {
+      data: {
+        sessionId,
```
jmooney-dd, line 105: suggestion: remove `sessionId` from data

sessionId will already be in log at this point as we append it earlier in the lambda

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+  }
+
+  // Check if the response has the expected structure
+  if (!parsedBody || typeof parsedBody.finish !== "string") {
+    logger.error(
+      LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_FAILURE,
+      {
+        data: {
+          parsedBody,
```
chris-cooksley-gds, line 137: I'm nervous about logging `parsedBody` here. A _valid_ biometric session would contain a lot of PII, and if we somehow got a malformed one through, there's no guarantee that the PII wouldn't be there (e.g. if they renamed the `finish` attribute unexpectedly but everything else was still there, we'd be logging a lot of PII here). Can we instead just indicate in the log data that there was an invalid response structure from ReadID?

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,157 @@
...
+  logger.debug(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_ATTEMPT, {
+    data: {
+      biometricSessionId,
+    },
+  });
+
+  const getBiometricSessionResult: Result<SuccessfulHttpResponse, HttpError> =
+    await sendHttpRequest(httpRequest, retryConfig);
+
+  if (getBiometricSessionResult.isError) {
```
chris-cooksley-gds, line 65: Nitpick: I think a helper function for this error handling would aid readability.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -38,19 +42,20 @@ export async function lambdaHandlerConstructor(
   if (validateSqsEventResult.isError) {
     return;
   }
-  const sessionId = validateSqsEventResult.value;
 
-  logger.appendKeys({ sessionId });
```
chris-cooksley-gds, line 43: Why has this been removed?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: This still needs updating.

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -311,6 +311,50 @@ export class LogMessage implements LogAttributes {
...
+
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR =
+    new LogMessage(
+      "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR",
+      "Error sending message to IPV Core",
+    );
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_TXMA_EVENT_ERROR = new LogMessage(
+    "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_TXMA_EVENT_ERROR",
+    "Error sending event to TxMA",
+  );
```
jmooney-dd, line 343: issue: we don't need this log message as we have a shared message for this `MOBILE_ASYNC_ERROR_WRITING_AUDIT_EVENT`

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Yeah, removed

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -311,6 +311,50 @@ export class LogMessage implements LogAttributes {
...
+  static readonly SESSION_NOT_FOUND = new LogMessage(
+    "MOBILE_ASYNC_SESSION_NOT_FOUND",
+    "Session not found in session registry",
+  );
```
jmooney-dd, line 347: issue: there is no issue biometric credential prefix here so this should either:
- Have a prefix
- Not have a prefix and be moved to the shared section of log messages

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Yes, extracted out

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
jmooney-dd: Potentially you meant biometricSessionId here, if so I think this should be appended when we get it as well

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,90 @@ export async function lambdaHandlerConstructor(
...
+
+    // Non-retryable error - send error to IPV Core
+    logger.error(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_NON_RETRYABLE_ERROR, {
+      data: {
+        sessionId,
+        error,
+      },
+    });
+
+    if (
```
chris-cooksley-gds, line 109: This conditional is unnecessary - `sessionAttributes` has already been validated and will always have these fields

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,90 @@ export async function lambdaHandlerConstructor(
...
+    });
+
+    if (writeEventResult.isError) {
+      logger.error(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_TXMA_EVENT_ERROR);
+    }
+
+    return;
+  }
+
+  const session = biometricSessionResult.value;
```
chris-cooksley-gds, line 153:
```suggestion
  const biometricSession = biometricSessionResult.value;
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,90 @@ export async function lambdaHandlerConstructor(
...
+    logger.info(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_NOT_READY, {
+      data: { finish: session.finish },
+    });
+    logger.info(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED);
+    throw new RetainMessageOnQueue(
+      `Biometric session not ready: ${session.finish}`,
+    );
+  }
 
+  // Session is ready, continue processing
```
chris-cooksley-gds, line 165:
```suggestion
  // Biometric session is ready, continue processing
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +79,90 @@ export async function lambdaHandlerConstructor(
...
   }
+  const viewerKey = viewerKeyResult.value;
+
+  const biometricSessionResult = await dependencies.getBiometricSession(
+    config.READID_BASE_URL,
+    sessionId,
+    viewerKey,
+  );
+
+  if (biometricSessionResult.isError) {
```
chris-cooksley-gds, line 90: For readability, it would be good to extract this error handling into one or more helper functions within this file.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,155 @@
...
+        data: {
+          getBiometricSessionResponse,
+          biometricSessionId,
+        },
+      },
+    );
+
+    return errorResult({
+      message: "Empty response body from ReadID",
+      isRetryable: true,
```
chris-cooksley-gds, line 105: See other comment @tmulberry-gds - can we also change this to `false`

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,155 @@
...
+  const getBiometricSessionResult: Result<SuccessfulHttpResponse, HttpError> =
+    await sendHttpRequest(httpRequest, retryConfig);
+
+  if (getBiometricSessionResult.isError) {
+    const error = getBiometricSessionResult.value;
+    const statusCode = error.statusCode ? Number(error.statusCode) : undefined;
+    const retryableStatusCodes = retryConfig.retryableStatusCodes;
+
+    const isRetryable =
+      !statusCode || (retryableStatusCodes?.includes(statusCode) ?? false);
```
chris-cooksley-gds, line 70:
```suggestion
      !statusCode || (retryableStatusCodes.includes(statusCode));
```

^ (we know that `retryableStatusCodes` is an array, and `includes` will always return a boolean)

backend-api/src/functions/asyncIssueBiometricCredential/issueBiometricCredentialConfig.ts
```diff
@@ -8,8 +8,10 @@ import { logger } from "../common/logging/logger";
 import { LogMessage } from "../common/logging/LogMessage";
 
 const REQUIRED_ENVIRONMENT_VARIABLES = [
+  "READID_BASE_URL",
```
jmooney-dd, line 11: issue: I don't think this lambda has access to this envVar as it's not global

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +420,119 @@ describe("Async Issue Biometric Credential", () => {
...
+          finish: "PROCESSING",
+        },
+      });
+    });
+
+    it("Throws RetainMessageOnQueue with appropriate message", () => {
+      expect(lambdaError).toBeInstanceOf(RetainMessageOnQueue);
+      expect((lambdaError as RetainMessageOnQueue).message).toMatch(
+        /Biometric session not ready: PROCESSING/,
+      );
```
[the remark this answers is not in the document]
tmulberry-gds: Comitted

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +420,119 @@ describe("Async Issue Biometric Credential", () => {
...
+            "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_NON_RETRYABLE_ERROR",
+        });
+      });
+
+      it("Sends error to IPV Core", () => {
+        expect(mockSuccessfulSendMessageToSqs).toHaveBeenCalledWith(
+          "mockIpvcoreOutboundSqs",
+          {
+            sub: expect.any(String),
+            state: expect.any(String),
```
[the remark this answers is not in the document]
tmulberry-gds: Updated to use values

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: I removed both of them

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Removed

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: Yeah, removed the parsedbody log

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+  }
+
+  // Check if the response has the expected structure
+  if (!parsedBody || typeof parsedBody.finish !== "string") {
+    logger.error(
+      LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_FAILURE,
+      {
+        data: {
+          validFinishProperty: parsedBody?.finish,
```
chris-cooksley-gds, line 131:
```suggestion
          invalidFinishProperty: parsedBody?.finish,
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -133,6 +195,30 @@ const handleGetSessionError = async (
   }
 };
 
+const handleSendMessageToOutboundQueue = async (
```
chris-cooksley-gds, line 198: This function only sends a server error message to the outbound queue (as opposed to a success message or some other kind of error). Could we either (a) make it suitably generic so that it can be re-used for other errors/non-errors, or (b) rename it to indicate that it only sends server errors?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -74,7 +83,60 @@ export async function lambdaHandlerConstructor(
...
+    const error: GetBiometricSessionError = biometricSessionResult.value;
+
+    // Check if the error was retryable based on error info
+    if (error.isRetryable) {
+      throw new RetainMessageOnQueue(
+        `Retryable error retrieving biometric session`,
+      );
+    }
+
+    handleSendMessageToOutboundQueue(dependencies, sessionAttributes, config);
```
chris-cooksley-gds, line 105: I think we need to `await` this since it's asynchronous.

backend-api/infra/functions/issueBiometricCredential.yaml
chris-cooksley-gds, line 1: Reminder (commenting here just to have somewhere to put it) - once the PR is ready for review, can we please test this in dev (against the readid mock) and attach evidence to the ticket to confirm that it works as expected? Will have to be a manual test for now - i.e. put a message in the vendor processing queue and show in the logs that we process it successfully

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -311,6 +311,42 @@ export class LogMessage implements LogAttributes {
...
+    new LogMessage(
+      "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_SUCCESS",
+      "Successfully retrieved biometric session from ReadID",
+    );
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_ATTEMPT =
+    new LogMessage(
+      "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_ATTEMPT",
+      "Retrieving biometric session from ReadID",
+    );
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_NOT_READY = new LogMessage(
```
chris-cooksley-gds, line 330:
```suggestion
  static readonly ISSUE_BIOMETRIC_CREDENTIAL_BIOMETRIC_SESSION_NOT_READY = new LogMessage(
```

backend-api/src/functions/common/logging/LogMessage.ts
```diff
@@ -311,6 +311,42 @@ export class LogMessage implements LogAttributes {
...
+    "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_NOT_READY",
+    "Biometric session not ready (finish status not DONE)",
+  );
+
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR =
+    new LogMessage(
+      "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_IPV_CORE_MESSAGE_ERROR",
+      "Error sending message to IPV Core",
+    );
+  static readonly ISSUE_BIOMETRIC_CREDENTIAL_RETRYABLE_ERROR = new LogMessage(
```
chris-cooksley-gds, line 340: Can these last two be deleted now?

backend-api/infra/functions/issueBiometricCredential.yaml
```diff
@@ -105,7 +105,9 @@ Resources:
                   - sqs:DeleteMessage
                   - sqs:GetQueueAttributes
                   - sqs:ChangeMessageVisibility
-                Resource: !GetAtt VendorProcessingSQS.Arn
+                Resource: 
+                  - !GetAtt VendorProcessingSQS.Arn
+                  - !GetAtt IPVCoreSQS.Arn
```
chris-cooksley-gds, line 110: I don't think this lambda needs to be able to perform these actions on the IPV core outbound queue (these are the actions you'd need to perform on a _source_ queue for a lambda). But the `AsyncAbortSessionFunctionIPVCoreSQSPolicy` should be a good model for what's needed to write to the outbound queue.

backend-api/infra/functions/issueBiometricCredential.yaml
chris-cooksley-gds, line 1: Thanks for the evidence added to the ticket. Some followups:
* Could we get evidence of a _successful_ write of the error message to the IPV outbound queue? Based on the permissions below I assume this won't work yet, so would be good to confirm that it does once the permissions are updated.
* Out of curiosity - how were you able to simulate ReadID returning `$.finish == 'NOT_DONE'`, or a retryable error? Is that something we can just set up with the ReadID mock? (I haven't used it properly before)

chris-cooksley-gds: On that last point, I notice in the logs that the retryable errors seem to be coming _after_ a logged successful response from ReadID, which I wouldn't have thought was possible given the code in `getBiometricSession`. What's up with that? 
image

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    jest.resetModules();
+  });
+
+  describe("On every call", () => {
+    beforeEach(async () => {
+      mockSendHttpRequest = jest.fn().mockResolvedValue(
+        successResult({
+          statusCode: 200,
+          body: JSON.stringify({
+            id: mockSessionId,
```
chris-cooksley-gds, line 50: Nitpick: can we remove the `id`s from these

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
```diff
@@ -0,0 +1,334 @@
...
+import { expect } from "@jest/globals";
+import "../../../../tests/testUtils/matchers";
+import { ISendHttpRequest } from "../../adapters/http/sendHttpRequest";
+
+describe("getBiometricSession", () => {
+  let result: Result<BiometricSession, GetBiometricSessionError>;
+  let consoleDebugSpy: jest.SpyInstance;
+  let consoleErrorSpy: jest.SpyInstance;
+  let mockSendHttpRequest: ISendHttpRequest;
+  const mockSessionId = "mockSessionId";
```
chris-cooksley-gds, line 16:
```suggestion
  const mockBiometricSessionId = "mockBiometricSessionId";
```

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,151 @@
...
+        data: {
+          error,
+          isRetryable,
+          biometricSessionId,
+        },
+      },
+    );
+
+    return errorResult({
+      statusCode,
```
chris-cooksley-gds, line 80: Still returning the statusCode here, which should no longer be needed.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
```diff
@@ -0,0 +1,334 @@
...
+      });
+
+      it("Logs error", () => {
+        expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+          messageCode:
+            "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_FAILURE",
+        });
+      });
+
+      it("Returns an error result with retryable flag", () => {
```
chris-cooksley-gds, line 244:
```suggestion
      it("Returns an error result with retryable flag set to false", () => {
```

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    it("Logs success at debug level", () => {
+      expect(consoleDebugSpy).toHaveBeenCalledWithLogFields({
+        messageCode:
+          "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_SUCCESS",
+      });
+    });
+
+    it("Returns successResult containing biometric session", () => {
+      expect(result).toEqual(successResult(mockSession));
+      expect(mockSendHttpRequest).toBeCalledWith(
```
chris-cooksley-gds, line 330: Nitpick: can be its own `it` block as it's an unrelated assertion.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +414,104 @@ describe("Async Issue Biometric Credential", () => {
...
+    });
+
+    it("Logs the appropriate message", () => {
+      expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
+        messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_NOT_READY",
+        data: {
+          finish: "PROCESSING",
+        },
+      });
+    });
```
chris-cooksley-gds, line 434: Can we test that it logs COMPLETED as well?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +528,46 @@ describe("Async Issue Biometric Credential", () => {
...
+
+    describe("Given sessionState is ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_FINISHED", () => {
       beforeEach(async () => {
         await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
       });
 
-      it("Passes correct arguments to get secrets", () => {
+      it("Passes correct arguments to get secrets", async () => {
+        await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
```
chris-cooksley-gds, line 560: Why has this been added here and in the test below?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +528,46 @@ describe("Async Issue Biometric Credential", () => {
...
         expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
           messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
-          sessionId: mockSessionId,
         });
       });
     });
 
-    describe("Given sessionState is ASYNC_BIOMETRIC_SESSION_FINISHED", () => {
+    it("Passes correct arguments to get biometric session", async () => {
+      await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
```
chris-cooksley-gds, line 539: Shouldn't need this (or the one below), since we already call it in the beforeEach

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +534,37 @@ describe("Async Issue Biometric Credential", () => {
...
+    });
+
+    it("Logs COMPLETED", async () => {
+      await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
+      expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
+        messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
       });
+    });
 
-      it("Passes correct arguments to get secrets", () => {
```
chris-cooksley-gds, line 402: Why is this test case being deleted?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -75,6 +84,74 @@ export async function lambdaHandlerConstructor(
...
+      getNowInMilliseconds: Date.now,
+      componentId: config.ISSUER,
+      ipAddress: undefined,
+      txmaAuditEncoded: undefined,
+      redirect_uri: sessionAttributes?.redirectUri,
+      suspected_fraud_signal: undefined,
+    });
+
+    if (writeEventResult.isError) {
+      logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT);
```
chris-cooksley-gds, line 133: Can we add `data` with the audit event name? (cf. what we log when we fail to write an event in biometricToken or abortSession)

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -370,6 +414,185 @@ describe("Async Issue Biometric Credential", () => {
...
+          it("Does not log COMPLETED", () => {
+            expect(consoleInfoSpy).not.toHaveBeenCalledWithLogFields({
+              messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
+            });
+          });
+        });
+
+        describe("Given writing TxMA event succeeds", () => {
+          beforeEach(async () => {
+            dependencies.getSessionRegistry = () => ({
```
chris-cooksley-gds, line 560: Why are we overriding the session registry to make the get operation fail, for this scenario?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +609,43 @@ describe("Async Issue Biometric Credential", () => {
         expect(mockGetSecretsSuccess).not.toHaveBeenCalled();
       });
 
-      it("Logs COMPLETED with sessionId", async () => {
+      it("Logs COMPLETED", async () => {
         expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
           messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
-          sessionId: mockSessionId,
```
chris-cooksley-gds, line 392: Why has this been removed?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +609,43 @@ describe("Async Issue Biometric Credential", () => {
...
           secretNames: ["mockBiometricViewerAccessKey"],
           cacheDurationInSeconds: 900,
         });
       });
 
-      it("Logs COMPLETED with sessionId", async () => {
+      it("Logs COMPLETED", async () => {
         expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
           messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
-          sessionId: mockSessionId,
```
chris-cooksley-gds, line 412: Why has this been removed?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -386,30 +609,43 @@ describe("Async Issue Biometric Credential", () => {
...
         });
       });
     });
 
-    describe("Given sessionState is ASYNC_BIOMETRIC_SESSION_FINISHED", () => {
+    it("Passes correct arguments to get biometric session", async () => {
```
chris-cooksley-gds, line 619: Should these two assertions be nested within `Given sessionState is ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_FINISHED`? We only get a biometric session in that scenario.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: I can see you've added the `writeEventResult` as log data. To clarify what I meant by this - we want it to match what we do elsewhere, e.g. [here](https://github.com/govuk-one-login/mobile-id-check-async/blob/main/backend-api/src/functions/asyncBiometricToken/asyncBiometricTokenHandler.ts#L282).

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -75,6 +84,76 @@ export async function lambdaHandlerConstructor(
     throw new RetainMessageOnQueue("Failed to retrieve biometric viewer key");
   }
 
+  const viewerKey = viewerKeyResult.value;
+
+  const biometricSessionResult = await dependencies.getBiometricSession(
+    config.READID_BASE_URL,
+    sessionAttributes.sessionId,
```
chris-cooksley-gds, line 91: This still needs changing. Please see unaddressed comment [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/586#discussion_r2048536820).

kikidawson-gds (dismissed):

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
tmulberry-gds: Removed Ids

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
```diff
@@ -0,0 +1,334 @@
...
+      });
+
+      it("Logs error", () => {
+        expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+          messageCode:
+            "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_FAILURE",
+        });
+      });
+
+      it("Returns an error result with retryable flag", () => {
```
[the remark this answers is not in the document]
tmulberry-gds: updated to say 'set to false'

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.test.ts
tmulberry-gds: updated to say 'set to false'

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Added it back in

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: This was the reason. Pushed changes

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Resolved

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: I removed the id dield

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
```diff
@@ -0,0 +1,146 @@
...
+  const retryConfig = {
+    retryableStatusCodes: [
+      429, 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, 511,
+    ],
+    delayInMillis: 50,
+  };
+
+  logger.debug(LogMessage.ISSUE_BIOMETRIC_CREDENTIAL_GET_FROM_READID_ATTEMPT, {
+    data: {
+      biometricSessionId,
```
chris-cooksley-gds, line 53: Nitpick (also applies below) - we've already appended the biometricSessionId to the logs in the handler, so this is redundant information.

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
chris-cooksley-gds: I think readability trumps consistency here (per our values/principles), but it's non-blocking - let's leave for a future refactor.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
chris-cooksley-gds: I'd lean towards extracting the full error handling for this scenario into a helper, too - but not a priority at this stage so we can leave this for a future refactor.

backend-api/infra/functions/issueBiometricCredential.yaml
chris-cooksley-gds: OK. I think ideally we'd want to test this through the readid mock, so that we know that we actually handle retryable failures over the network in the way we expect. But, I'm not sure how easy that is (or if it's even possible), so I'm happy with this workaround for now.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds: Github's showing this as being removed, still 
image

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds: Github's showing this as being removed, still. 
image

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds: My comment wasn't clear - I think the scenarios highlighted in this screenshot _should_ be nested within `Given sessionState is ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_FINISHED`. 

Currently, there are only two valid scenarios (session state either RESULT_SENT or ISSUE_BIOMETRIC_CREDENTIAL_FINISHED) and I think every test case should therefore be nested within one of those, rather than 'loose' at the top level.

chris-cooksley-gds: image

backend-api/src/functions/asyncIssueBiometricCredential/getBiometricSession/getBiometricSession.ts
tmulberry-gds: I have removed this from the log files

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Weird, it didnt push the changes

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Same as above, didn tpush the changes for some reason

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds: This one is still missing

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -391,181 +390,201 @@ describe("Async Issue Biometric Credential", () => {
...
             messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
-            sessionId: mockSessionId,
           });
         });
       });
     });
   });
 
-  describe("When there is an error getting secrets", () => {
+  describe("When biometric session state is RESULT_SENT", () => {
```
chris-cooksley-gds, line 399:
```suggestion
  describe("When session state is ASYNC_RESULT_SENT", () => {
```

This is the user session, not the biometric session.

As a reminder:
* the user session is stored in dynamo, managed by us, has a bunch of useful fields in (e.g. subject ID, client ID, redirect URI), and follows a state machine (e.g. ASYNC_AUTH_SESSION_CREATED, ASYNC_BIOMETRIC_TOKEN_ISSUED, etc)
* the biometric session is managed by ReadID. The mobile app started the biometric session with ReadID and we're retrieving it in this ticket. It will include results of the checks that ReadID has done.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -391,181 +390,201 @@ describe("Async Issue Biometric Credential", () => {
...
+    it("Logs COMPLETED with sessionId", async () => {
+      expect(consoleInfoSpy).toHaveBeenCalledWithLogFields({
+        messageCode: "MOBILE_ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_COMPLETED",
+        sessionId: mockSessionId,
+      });
     });
   });
 
-  describe("When biometric session is not ready", () => {
+  describe("When biometric session state is ASYNC_ISSUE_BIOMETRIC_CREDENTIAL_FINISHED", () => {
```
chris-cooksley-gds, line 422:
```suggestion
  describe("When session state is ASYNC_BIOMETRIC_SESSION_FINISHED", () => {
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: I've split them up into their corrosponding pths

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: commited this

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: comitted this

chris-cooksley-gds (approved):

PR #617, opened 2025-05-06
DCMAW-12524: Send the DCMAW_ASYNC_CRI_END event to TxMA

tmulberry-gds:
## Jira Ticket

DCMAW-12524

## Description of changes

- Added `DCMAW_ASYNC_CRI_END` event type
- Created `handleSendCriEvent` method for sending CRI event

## Review guidance
<details>
<summary>Review checklist</summary>

### Functional Review
- **Functionality**: Does it meet the acceptance criteria on the ticket and work as expected?
- **Requirements**: Does the code meet functional and non-functional requirements including compliance with programme standards and security gates?

### Security & Compliance
- **Personally Identifiable Information**: Is it possible for PII to be logged?
- **Security Considerations**: Are there any security implications that need to be addressed?

### Quality Assurance
- **Testing**: Is the code well-tested with sufficient coverage to provide confidence in correctness?
- **Edge Cases**: Have edge cases been considered and handled appropriately?

### Code Quality
- **Readability**: Is the code easy to understand for all team members, with clear naming and appropriate documentation?
- **Maintainability**: Is the code easy to change, reuse, and extend?
- **Code Style**: Does it follow our coding conventions and best practices?
- **Code Quality**: Is the code maintainable and following best practices? See [Values, Principles & Practices](https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/4955865126/ID+Check+Web+-+Values+Principles+and+Practices)

### Observability & Operations
- **Observability**: Are there appropriate logs/metrics that would help debug and monitor the service?
- **Performance**: Are there any performance considerations or potential bottlenecks?

### Documentation
- **Documentation**: Is the code well documented? Is there any existing documentation that needs updating?
- **Comments**: Are complex sections of code adequately commented if the intent is not clear?

### Review PR:
- **Title**: Contains ticket number and clear summary of change
- **Description**: Has clear description of change
</details>


backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -248,6 +248,12 @@ export async function lambdaHandlerConstructor(
     return;
   }
 
+  await handleSendCriEvent({
+    eventService,
+    sessionAttributes,
+    issuer: config.ISSUER,
+  });
```
tmulberry-gds, line 255: Is this the correct place for this? In v1 it's sitting around this area so unsure if this is correct

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -496,3 +502,38 @@ export const buildCredentialJwtPayload = (jwtData: {
...
+    txmaAuditEncoded: undefined,
+    redirect_uri: redirectUri,
+    suspected_fraud_signal: undefined,
+  });
+
+  if (writeEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: { auditEventName: "DCMAW_ASYNC_CRI_5XXERROR" },
+    });
+  }
```
tmulberry-gds, line 536: Dont know if we want to have the error log within the method. Thoughts?

jmooney-dd: Would the alternative be where you call this function?

I think the log should be where you have it currently, as close to the failure as pos

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1080,6 +1080,41 @@ describe("Async Issue Biometric Credential", () => {
         );
       });
 
+      describe("Given sending DCMAW_ASYNC_CRI_END event fails", () => {
```
jmooney-dd, line 1083: issue: the way the happy path describe block is currently structured means that multiple beforeEaches will fire which we don't want

You could fix this by structuring the happy path block like

describe("Happy path")
REMOVE THIS BEFORE EACH
  describe("Given writing CRI end fails)
    beforeEach
  describe("Given writing Cri end suceeds")
    beforeEach

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -496,3 +502,36 @@ export const buildCredentialJwtPayload = (jwtData: {
     vc: credential,
   };
 };
+
+interface HandleSendCriEndEventData {
+  eventService: IEventService;
+  sessionAttributes: SessionAttributes;
+  issuer: string;
+}
+const handleSendCriEndEvent = async (
```
jmooney-dd, line 511: nitpick (entirely preference): this is just writing and logging a if it fails, which is the entire process of sending an event, so we could rename this to just `sendCriEndEvent`

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -496,3 +502,36 @@ export const buildCredentialJwtPayload = (jwtData: {
     vc: credential,
   };
 };
+
+interface HandleSendCriEndEventData {
+  eventService: IEventService;
+  sessionAttributes: SessionAttributes;
+  issuer: string;
```
jmooney-dd, line 509: note: `issuer` is in the `sessionAttributes` so you could use that value instead of having an extra param. unsure how people feel about that..

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
jmooney-dd: You've placed it directly after the result sent update session call so I think this is correct. It shouldn't come before that call because the sessionAttributes would show the wrong session state

jmooney-dd: I think it still makes sense to be where you've put it but may we worth getting other opinions

backend-api/src/functions/services/events/types.ts
```diff
@@ -50,6 +50,7 @@ export type GenericEventNames =
   | "DCMAW_ASYNC_CRI_4XXERROR"
   | "DCMAW_ASYNC_CRI_5XXERROR"
   | "DCMAW_ASYNC_CRI_ERROR"
+  | "DCMAW_ASYNC_CRI_END"
```
jmooney-dd, line 53: issue: you are missing the test for this in the event service

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Updated the structure of the tests

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Yeah I agree, renamed to sendCriEndEvent

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Removed issuer param and used sessionAttributes

backend-api/src/functions/services/events/types.ts
tmulberry-gds: Added test

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: Agree, I like having related operations within specific methods

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
jmooney-dd: I think we've now got an extra one that hasn't been removed at the top. So all the its should go in the describe blocks illustrated above.

Happy to chat thru if easier

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1016,74 +1016,175 @@ describe("Async Issue Biometric Credential", () => {
...
+          expect(mockSendMessageToSqsSuccess).toHaveBeenCalledWith(
+            "mockIpvcoreOutboundSqs",
+            {
+              "https://vocab.account.gov.uk/v1/credentialJWT": [mockSignedJwt],
+              state: mockClientState,
+              sub: mockSubjectIdentifier,
+            },
+          );
+        });
+        it("Logs the DCMAW_ASYNC_CRI_5XXERROR event failure", () => {
```
jmooney-dd, line 1086:
```suggestion

        it("Logs the DCMAW_ASYNC_CRI_5XXERROR event failure", () => {
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1016,74 +1016,175 @@ describe("Async Issue Biometric Credential", () => {
...
+          expect(consoleErrorSpy).toHaveBeenCalledWithLogFields({
+            messageCode: "MOBILE_ASYNC_ERROR_WRITING_AUDIT_EVENT",
+            data: {
+              auditEventName: "DCMAW_ASYNC_CRI_5XXERROR",
+            },
+          });
+        });
+        it("Logs COMPLETED with sessionId", () => {
```
jmooney-dd, line 1094:
```suggestion

        it("Logs COMPLETED with sessionId", () => {
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1016,74 +1016,175 @@ describe("Async Issue Biometric Credential", () => {
...
+          expect(mockSendMessageToSqsSuccess).toHaveBeenCalledWith(
+            "mockIpvcoreOutboundSqs",
+            {
+              "https://vocab.account.gov.uk/v1/credentialJWT": [mockSignedJwt],
+              state: mockClientState,
+              sub: mockSubjectIdentifier,
+            },
+          );
+        });
+        it("Calls writeGenericEvent with correct parameters", () => {
```
jmooney-dd, line 1169:
```suggestion

        it("Calls writeGenericEvent with correct parameters", () => {
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -496,3 +501,38 @@ export const buildCredentialJwtPayload = (jwtData: {
...
+    getNowInMilliseconds: Date.now,
+    ipAddress: undefined,
+    txmaAuditEncoded: undefined,
+    redirect_uri: redirectUri,
+    suspected_fraud_signal: undefined,
+  });
+
+  if (writeEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: { auditEventName: "DCMAW_ASYNC_CRI_5XXERROR" },
```
jmooney-dd, line 535: issue: wrong name

```suggestion
      data: { auditEventName: "DCMAW_ASYNC_CRI_END" },
```

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
chris-cooksley-gds, line 1094: Question - should we be logging completion when we fail to send this event? The log message indicates we have 'completed without issue' and I feel like this is an issue. Perhaps another way to look at it is - if we accidentally deployed a change that specifically broke the sending of this event, would we want to trigger a canary rollback?

Invoking @sandymay-dd for this as I'm not 100% sure, and because it arguably ties into some of the thinking around alarms for these failures.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1016,74 +1016,175 @@ describe("Async Issue Biometric Credential", () => {
     });
 
     describe("Happy path", () => {
-      beforeEach(async () => {
-        // Uses the default mockSessionRegistrySuccess which returns validBiometricSessionFinishedAttributes
-        await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
-      });
+      describe("Given sending DCMAW_ASYNC_CRI_END event fails", () => {
```
chris-cooksley-gds, line 1019: Nitpick: this scenario isn't part of the happy path IMO (it's happier than some of the other errors, but it's still an error). Can we extract this describe block out to sit above (and on the same level of nesting as) `describe("Happy path"...`?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1016,74 +1016,175 @@ describe("Async Issue Biometric Credential", () => {
...
+            getEventService: () => mockFailingEventService,
+          };
 
-      it("Passes correct arguments to get secrets", () => {
-        expect(mockGetSecretsSuccess).toHaveBeenCalledWith({
-          secretNames: ["mockBiometricViewerAccessKey"],
-          cacheDurationInSeconds: 900,
+          await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
+        });
+        it("Passes correct arguments to get secrets", () => {
```
chris-cooksley-gds, line 1028: Nitpick: IMO (and following convention we've established elsewhere) we don't need to repeat all of the 'Passes correct arguments' type assertions in the error scenarios, so long as we've got them in the core happy path scenarios. It's a lot of repetition that doesn't give us a lot of extra confidence, and will make it a bit harder to update e.g. the interfaces of these dependencies having this in lots of places.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
sandymay-dd: >  if we accidentally deployed a change that specifically broke the sending of this event, would we want to trigger a canary rollback

We would want to roll back if this event fails to send. Therefore we want to `return` without this log. To do this, I assume the error handling would be inline in the lambda - similar to how we handle `const updateSessionResult` just above this line in the lambda.

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
jmooney-dd: Yeah I agree. We discussed these blocks earlier when fixing the different level beforeEaches and ended up duplicating those but no need if we're doing it on the 'happy' path

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
jmooney-dd: I think given this [discussion](https://github.com/govuk-one-login/mobile-id-check-async/pull/617#discussion_r2075784866) this suggestion makes more sense but we did discuss this earlier and I thought that this was a happy path scenario as the ticket says log and delete message from queue, implying we'd then log success which to me is happy

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Removed the duplicate tests in the unhappy happy path

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
tmulberry-gds: @chris-cooksley-gds 

Your [comment](https://github.com/govuk-one-login/mobile-id-check-async/pull/617#discussion_r2075784866) would be resolved by moving this out of the method and adding in a return, thoughts?

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: I've rejigged the code a bit and extracted out some logic, i.e removed the method itself and inlined it to the lambda a la @sandymay-dd

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1015,75 +1015,116 @@ describe("Async Issue Biometric Credential", () => {
...
 
-      it("Passes correct arguments to get biometric session", () => {
-        expect(mockGetBiometricSessionSuccess).toHaveBeenCalledWith(
-          "mockReadIdBaseUrl",
-          mockBiometricSessionId,
-          "mockViewerKey",
-        );
-      });
+    describe("Happy path", () => {
+      describe("Writes DCMAW_ASYNC_CRI_END event to TxMA", () => {
```
jmooney-dd, line 1039: suggestion: remove this describe block, we have the it below that says we write to txma

I would also remove the re assignment of the dependencies in the beforeEach and just have the call to the lambda because:
- No need to actually re assign as we assign it the success mock at the top of the file
- I don't think it's worth being declarative about the event service dep only here, it would make sense to then declare all of the success deps here and I don't think it's worth personally

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1015,75 +1015,116 @@ describe("Async Issue Biometric Credential", () => {
...
+        beforeEach(async () => {
+          dependencies = {
+            ...dependencies,
+            getEventService: () => mockSuccessfulEventService,
+          };
```
jmooney-dd, line 1044: suggestion: remove this describe block:
- we have the it below that says we write to txma
- It's actually just covering all the happy path tests anyway

I would also remove the re assignment of the dependencies in the beforeEach and just have the call to the lambda because:
- No need to actually re assign as we assign it the success mock at the top of the file
- I don't think it's worth being declarative about the event service dep only here, it would make sense to then declare all of the success deps here and I don't think it's worth personally

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
```diff
@@ -248,6 +248,28 @@ export async function lambdaHandlerConstructor(
...
+    redirect_uri: redirectUri,
+    suspected_fraud_signal: undefined,
+  });
+  if (writeCriEndEventResult.isError) {
+    logger.error(LogMessage.ERROR_WRITING_AUDIT_EVENT, {
+      data: { auditEventName: "DCMAW_ASYNC_CRI_END" },
+    });
+    return;
+  }
+
```
jmooney-dd, line 272: question: previously I think you went with a helper func to slim down the lambda, are we not still able to do this? 

I think that helper function might be used as well for the next PR as well?

tmulberry-gds: I did go with a helper func, but to resolve an issue [here](https://github.com/govuk-one-login/mobile-id-check-async/pull/617#discussion_r2075784866), I needed to extract out the logger so as to not over complicate the function, leaving just the eventService call which I dont think really needed its own function, so I got rid of it all together. 

In the next part of the ticket, if it makes sense to move it around again that I'll do that

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Oh yeah good point. Removed it :)

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
jmooney-dd: In the helper function are you still not able to:

- If the audit event fails
  - Log failure
  - return
- If the the audit event succeeds don't return

Am probs missing something sorry!

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
```diff
@@ -1015,9 +1015,28 @@ describe("Async Issue Biometric Credential", () => {
...
+      beforeEach(async () => {
+        dependencies = {
+          ...dependencies,
+          getEventService: () => mockFailingEventService,
+        };
+
+        await lambdaHandlerConstructor(dependencies, validSqsEvent, context);
+      });
+
+      it("Logs the DCMAW_ASYNC_CRI_END event failure", () => {
```
jmooney-dd, line 1028: suggestion: Add a does not log completed test here

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.ts
jmooney-dd: Ah when I add that test this DOES NOT work

jmooney-dd: Happy to see how this evolves then on the next PR

backend-api/src/functions/asyncIssueBiometricCredential/asyncIssueBiometricCredentialHandler.test.ts
tmulberry-gds: Added the test

jmooney-dd (approved):

PR #1169, opened 2026-05-01
Add amazonq directory to .gitignore

sandymay-dd:


.gitignore
```diff
@@ -31,3 +31,6 @@ parameters-*-cf-dist.json
 # files generated when running the sts-mock jwks helper script
 jwks.json
 private-key.json
+
+# Amazon Q 
+.amazonq/
```
phad, line 36: I'm not sure if the trailing `/` is needed?  We don't use it consistently in this file.
(e.g. `node_modules`, `.vscode` are definitely directories.)

phad (dismissed):
nitty comment, but lgtm

phad (dismissed):
lgtm

.gitignore
sandymay-dd: If the trailing `/` is missing then a file called `.amazonq` would be git ignored. I was only intending to ignore the directory but I can update this to also ignore a file called `.amazonq` if we wished by removing the `/`?

